PR #323, opened 2023-05-18
Lightning v2

joao-alex-cunha:
NOT TO BE LANDED YET

Adding support for Lightning 2.+

Removes most changes needed in the custom fork being used for IPUs.
Training with the model weights and data in FP 16 precision is referred in Lightning as "16-true".


DomInvivo (approved):
Let's just fix the CI, but everything is perfect

Issue #255, opened 2023-03-20
Dependencies review and warnings


hadim:
_Original comment at https://github.com/valence-discovery/goli/pull/254#issuecomment-1476819287_

I would suggest reviewing the goli dependencies (less deps is always better). Here are a few suggestions/questions:

- `tensorboard`: seems to be unused?
- `torchvision` is used in `ipu_simple_lightning.py` but not present in the env.yml file
- `hydra-core` seems to be unused?
- `optuna`: does not seem to be imported (should we remove it from the conda package?)
- `umap-learn`: seems unused?
- `torchmetrics` is set to `torchmetrics >=0.7.0,<0.11`: it would be nice to catch up with the latest versions, so we can remove the upper bound from the package (upper bounds tend to create dep conflicts in the long term)

I also spotted some warnings during the tests (you should see them as well). For any warnings related to goli, it would be nice to address them whenever it's possible.

Generally, it's important to separate the deps that are specific to the Python goli package from the deps used during the experiments. It's always best to minimize the number of dependencies in the package.

Another question: the goli package contains a QM9 dataset of 30MB. It's not that large, but it increases the total size of the package. Do you want to keep it there?

PR #343, opened 2023-06-06
Fix unittests

DomInvivo:
Checklist:

- [ ] Was this PR discussed in a issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR.
- [ ] Add tests to cover the fixed bug(s) or the new introduced feature(s) (if appropriate).
- [ ] Update the API documentation is a new function is added or an existing one is deleted.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ mamba env create -f env.yml -n graphium
 
 # Install Graphium in dev mode
 mamba activate graphium
-pip install -e .
+pip install --no-deps -e .
 ```
 
 ### For IPU developers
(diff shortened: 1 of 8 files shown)


hadim:
Thanks Dom. I think there was an issue with the ZINC-micro.zip file on the bucket. I fixed it, and it should do the job here.

Using zip is probably not the best here. For dataframe I recommend simply using parquet everywhere (types are preserved and compression comes for free + random access at the column level). For more complex datastructure, I strongly recommend using [zarr](https://zarr.readthedocs.io/en/stable/) (which is basically hdf5 but more modern).

hadim:
Also I think that ticket is still relevant: https://github.com/datamol-io/graphium/issues/255. The tests show 371 warnings. I recommend hiding the ones that we can't control (in the `pyproject.toml` file) and addressing the other ones. It's best to be proactive here.

PR #349, opened 2023-06-07
Zhiyil/some changes

zhiyil-graphcore:
Checklist:

- [x] modify config for ckpt so that last ckpt is saved
- [x] fix a few bugs about metrics in code and config
- [x] add validation only script
- [x] add baseline configs for small dataset
- [x] largemix bug fixes
- [x] largemix configs: multi-task and single-task 
---


graphium/trainer/predictor.py
```diff
@@ -565,9 +565,9 @@ def validation_epoch_end(self, outputs: Dict[str, Any]):
         concatenated_metrics_logs = self.task_epoch_summary.concatenate_metrics_logs(metrics_logs)
         concatenated_metrics_logs["val/mean_time"] = self.mean_val_time_tracker.mean_value
         concatenated_metrics_logs["val/mean_tput"] = self.mean_val_tput_tracker.mean_value
-
-        lr = self.optimizers().param_groups[0]["lr"]
-        concatenated_metrics_logs["lr"] = lr
+        if hasattr(self.optimizers(), "param_groups"):
+            lr = self.optimizers().param_groups[0]["lr"]
+            concatenated_metrics_logs["lr"] = lr
```
callumm-graphcore, line 570: Why was this added?

zhiyil-graphcore: This is when doing validation only, we don't have `param_groups` attr for` self.optimizers()`

graphium/trainer/losses.py
```diff
@@ -9,7 +9,7 @@
 class HybridCELoss(_WeightedLoss):
     def __init__(
         self,
-        n_brackets,
+        n_brackets: int = 5,
```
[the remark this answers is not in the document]
zhiyil-graphcore: I didn't find where we can pass the initialization paramter to the loss function from the config. We can alternatively pass this in the `ipu_losses.py` .
```
  loss_fun:
    l1000_vcap: hybrid_ce_ipu
    l1000_mcf7: hybrid_ce_ipu
    pcba_1328: bce_ipu
    pcqm4m_g25: mae_ipu
    pcqm4m_n4: mae_ipu
```

expts/neurips2023_configs/config_small_gated_gcn.yaml
```diff
@@ -295,10 +295,12 @@ metrics:
   tox21:
     - name: auroc_flat
       metric: auroc_ipu
+      task: binary
```
DomInvivo, line 298: The `multitask_handlig` should change to `mean-per-label` if we want to match the litterature

graphium/data/dataset.py
```diff
@@ -368,6 +368,8 @@ def __getitem__(self, idx):
             data_dict = self.load_graph_from_index(idx)
             datum["features"] = data_dict["graph_with_features"]
             # these type changes are temporary for ipu dtype match
+            if hasattr(data_dict["labels"], "graph_l1000_mcf7"):
```
DomInvivo, line 371: okay for the experiments now. But please don't introduce this kind of hard-coding into the final code.

graphium/ipu/ipu_losses.py
```diff
@@ -107,6 +107,17 @@ def forward(self, input: Tensor, target: Tensor) -> Tensor:
 
 
 class HybridCELossIPU(HybridCELoss):
+    def __init__(
+        self,
+        n_brackets: int = 5,
```
DomInvivo, line 112: Remove default value for `n_brackets`. Pass it in the config file under the metric

DomInvivo (commented):
Few comments

expts/neurips2023_configs/config_small_gated_gcn.yaml
zhiyil-graphcore: Do we wanna change all the auroc_flat to `mean-per-label` or just for the tox21 case?

graphium/ipu/ipu_losses.py
zhiyil-graphcore: This is for the loss_func, is it happy to accept something like this?
```
   loss_fun:
    l1000_vcap: hybrid_ce_ipu
        n_brackets: 5
    l1000_mcf7: hybrid_ce_ipu
        n_brackets: 5
```

DomInvivo: We need to modify the loss_fun to work like the metrics and take parameters. I thought we did that a while ago, but it seems not.

DomInvivo:
```yaml
l1000_cap:
  name: hybrid_ce_ipu
  n_brackets: 5
```

zhiyil-graphcore: This above works fine, closing this conversation.

graphium/trainer/predictor.py
```diff
@@ -310,13 +310,17 @@ def _general_step(self, batch: Dict[str, Tensor], step_name: str, to_cpu: bool)
         }
         # preds = {k: preds[ii] for ii, k in enumerate(targets_dict.keys())}
         for task, pred in preds.items():
+            if pred.dtype == torch.float16:
+                pred_fp32 = pred.to(torch.float32)
+                pred = pred_fp32
+                preds[task] = pred_fp32
```
DomInvivo, line 316:
```python
preds[task] = pred.to(torch.float32)
```

graphium/trainer/predictor.py
```diff
@@ -310,13 +310,17 @@ def _general_step(self, batch: Dict[str, Tensor], step_name: str, to_cpu: bool)
...
             task_specific_norm = self.task_norms[task] if self.task_norms is not None else None
             if step_name != "train":
                 # apply denormalization for val and test predictions for correct loss and metrics evaluation
                 # targets for val and test were not normalized
                 # train loss will stay as the normalized version
                 preds[task] = task_specific_norm.denormalize(pred)
-            targets_dict[task] = targets_dict[task].to(dtype=pred.dtype)
+                targets_dict[task] = targets_dict[task].to(dtype=pred.dtype)
```
DomInvivo, line 323: Not sure it makes sense. Why not casting the dtype also at training?? It will crash if the targets and preds are not the same dtype.

Above, you convert to float32, but what if the targets are float16? or float64?

graphium/trainer/predictor_summaries.py
```diff
@@ -388,6 +388,10 @@ def update_predictor_state(
             task_losses: the task losses
             n_epochs: the number of epochs
         """
+        for task, pred in predictions.items():
+            if pred.dtype == torch.float16:
+                pred_fp32 = pred.to(torch.float32)
+                predictions[task] = pred_fp32
```
DomInvivo, line 394: Why such hard-coding? And what if the preds are float64?

DomInvivo (commented):
Few comments, seems a bit hacky and too much hard-coded

zhiyil-graphcore:
> All the changes look good. But there are way too many config files, and they're all copies of each other. This will polute the Repo, especially since it's now public. It is also prone to copy-paste errors and require all the files to change every time there's a code change.
> 
> I would request to keep only 1 full config file. For the others, create only a very simple configuration that will overwrite the specific path of the Yaml.
> 
> You can use the function [merge_dicts](https://github.com/datamol-io/graphium/blob/2790af4d95e8924671cc666fac002fd003130618/graphium/config/_loader.py#LL483C9-L483C9) and modify it with the option to not raise an error if a value already exists.
> 
> Alternatively, you can define bash files with [hydra](https://hydra.cc/docs/advanced/hydra-command-line-flags/) to manage changes between runs.

Yeah that's a very good point. Actually we don't have any results for the largemix and those single tasks yet, I wonder if you are happy if we now only show the toxymix multitask baseline and largemix multitask configs so that people could reproduce the toymix baseline results. For the single task ones, we could leave them out at the moment and add the configs in using the merg_dict method you mentioned above once the results are ready.

zhiyil-graphcore:
Same changes here: https://github.com/datamol-io/graphium/pull/362

PR #350, opened 2023-06-07
Removing hydrogens

s-maddrellmander:
Checklist:
- [ ] Waiting on tests to run to ensure this is working in the full dataset processing

- [ ] Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR.
- [ ] Add tests to cover the fixed bug(s) or the new introduced feature(s) (if appropriate).
- [ ] Update the API documentation is a new function is added or an existing one is deleted.

---


graphium/data/datamodule.py
```diff
@@ -982,6 +993,11 @@ def prepare_data(self):
 
             logger.info(f"Prepare single-task dataset for task '{task}' with {len(df)} data points.")
 
+            logger.info("Filtering the molecules for Hydrogen")
```
DomInvivo, line 996: Why not just pass `explicit_H: false` to the `featurizer` options in the yaml?

zhiyil-graphcore: it is false in the config: `explicit_H: False # if H is included` I think this is for hydrogen only molecules like H2, I wonder if molecules like this would be filtered out completely if we use the flag in the featurizer? And what happens with the graph indices and features etc if they do get filtered out?

s-maddrellmander: Yeah we end up with an empty molecule on removing the hydrogens from H2, which doesn't make a huge amount of sense and causes breaking issues with the dataset processing. And I wanted to make sure it would be caught under different formatting of the string.

hadim: Sorry to jump a little bit from nowhere here. I am working with Dom for context!

You are using a string-based replacement to remove the hydrogens which can be problematic for certain smiles (leading to invalid molecules): `smiles_without_h = re.sub('H', '', smiles)`.

A more robust approach could be to remove the implicit hydrogens from the molecule object directly. All the chemoinformatics software supports that option. With datamol you can use: `mol_no_hs = dm.remove_hs(mol)`.

Last, also note that you might not want to remove any explicit hydrogens since they could lead to a different molecule. I suggest you check your dataset after and before processing the explicit hydrogens and compare the inchikey or the hash of the molecule.

DomInvivo: Sorry I messed up! The problem is because of an error when processing the dataset!! I removed all Hs from the labels instead of only explicit ones.

Let's change this to throw away every molecule where the number of atoms is different then the number of node labels, or when the number of edges is different then the number of edge labels, or that the shape of a nodepair label is different then [num_atoms, num_atoms]

DomInvivo: A possible way is to populate the `idx_none` so that it is taken care by `self._filter_none_molecules`.

Since here you're only at the smiles level, then you don't know how many atoms and edge there are. So perhaps it's better to do it once it was featurized??

DomInvivo: @s-maddrellmander let me know if you need help with this issue

PR #352, opened 2023-06-08
fix IPU install instructions

joao-alex-cunha:
Fixes IPU installation instructions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
     { name = "Dominique Beaini", email = "dominique@valencediscovery.com" },
 ]
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.8"
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -10,7 +10,7 @@ authors = [
     { name = "Dominique Beaini", email = "dominique@valencediscovery.com" },
 ]
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.8"
```
DomInvivo, line 13: Change conda version as well??

DomInvivo: Done

hadim: the env.yml file is only used in that repo. you can ofc put >=3.8 (I just used >=3.10 because i saw it somewhere in the old goli so i thought it was the minimum python version used).

@DomInvivo you can also ask the CI to test for multiple python versions in a matrix CI at https://github.com/datamol-io/graphium/blob/main/.github/workflows/test.yml#L19.

hadim: also add it at https://github.com/datamol-io/graphium/blob/main/pyproject.toml#L25-L28

PR #364, opened 2023-06-22
Hybrid ce loss changes

zhiyil-graphcore:
## Changelogs

---

_Checklist:_

---

Fix to hybridceloss, but it is not actually helping the hybridceloss for val and test using the debug config.


graphium/trainer/losses.py
```diff
@@ -59,10 +60,14 @@ def forward(self, input: Tensor, target: Tensor) -> Tensor:
         """
 
         target = target.flatten()
-
-        regression_input = torch.inner(input, self.brackets.to(input.device))
+        # regression loss needs normalized logits to probability as input to do inner product with self.brackets
+        # we apply softmax on the raw logits first
+        softmax_input = self.softmax(input)
```
DomInvivo, line 65: I thought we discussed that the `softmax` should go as the `last_activation` since we want the classification metrics to work

DomInvivo (changes requested):

graphium/trainer/losses.py
zhiyil-graphcore: I mentioned about this on [slack](https://graphcore.slack.com/archives/C03NG9SA3QC/p1687515389789519) about the auroc and average precision metrics. It looks like they are happy with both probabilities or logits. In that case we don't need a custom CE loss? Or do you think there would be other classification metrics which only accept probabilities as the input?
```
`    Accepts the following input tensors:

    - ``preds`` (float tensor): ``(N, C, ...)``. Preds should be a tensor containing probabilities or logits for each
      observation. If preds has values outside [0,1] range we consider the input to be logits and will auto apply
      softmax per sample.`
```

DomInvivo: Thanks, just saw your Slack message!

I checked also Accuracy and f1. I think this is consistent across metrics 🎉

PR #367, opened 2023-06-27
some fixes: keys() and use normalization flag on val and test for baseline configs 

zhiyil-graphcore:
## Changelogs

---

_Checklist:_

---


callumm-graphcore:
I would strongly suggest that we make the `keys` vs `keys()` thing for the PyG batches a function, rather than just making a bunch of independent changes

graphium/data/collate.py
```diff
@@ -124,7 +124,7 @@ def collage_pyg_graph(pyg_graphs: Iterable[Union[Data, Dict]], batch_size_per_pa
 
     pyg_batch = []
     for pyg_graph in pyg_graphs:
-        for pyg_key in pyg_graph.keys:
+        for pyg_key in pyg_graph.keys() if isinstance(pyg_graph, object) else pyg_graph.keys:
```
DomInvivo, line 127: `object`??? Isn't everything in python an object? Abd aren't the `pyg_graph` always `pyg.data.Data` objects or `GraphDict` objects? So why being so generic? 

If it's the `GraphDict` object that's causing the issue, you can follow @callumm-graphcore suggestion and do this, although it might break some other code that check for `GraphDict` objects to use `keys()`:

```python
class GraphDict(dict):
...
    @property
    def keys(self):
        return list(super().keys())

    @property
    def values(self):
        return list(super().self.values())
```

zhiyil-graphcore: Yeah you're absolutely right, I was not aware that list and dict are objects as well. I've adopted your method and @callumm-graphcore 's method on this.

expts/neurips2023_configs/baseline/config_small_gin_baseline.yaml
```diff
@@ -65,6 +65,7 @@ datamodule:
         seed: *seed
         task_level: graph
         label_normalization:
+          normalize_val_test: True
```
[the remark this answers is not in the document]
zhiyil-graphcore: Sure, I've made the changes.

DomInvivo (approved):

Issue #365, opened 2023-06-26
Failing unit tests cuz OGB has changed its URL for freesolv


hadim:
See the CI logs: https://github.com/datamol-io/graphium/actions/runs/5373722733/jobs/9748378988

```
FAILED tests/test_datamodule.py::Test_DataModule::test_caching - ValueError: The file being copied does not exist: http://snap.stanford.edu/ogb/data/graphproppred/csv_mol_download/freesolv.zip
FAILED tests/test_datamodule.py::Test_DataModule::test_ogb_datamodule - ValueError: The file being copied does not exist: http://snap.stanford.edu/ogb/data/graphproppred/csv_mol_download/freesolv.zip
```

hadim:
ping @DomInvivo

DomInvivo:
The Stanford servers are down
https://cs.stanford.edu/srcf_404/

Let's ignore that for now and circle back in a few days.

![image](https://github.com/datamol-io/graphium/assets/47570400/29fa0cf4-6d5b-4f88-8eff-f059096092ba)

DomInvivo:
The URL is now working

PR #370, opened 2023-06-28
Tutorial update and dependency fix 

shenyangHuang:
This PR has a temporary fix for the IPU dependency issue. The installation on the [README](https://github.com/datamol-io/graphium/blob/dependency/README.md) has been updated however the lightning version remains too old. 

All current tutorials are able to run and tested. 
Removed older tutorials: 
- making gnn networks (for dgl only)
- using gnn layers (for dgl only) 
- config ipu tutorial (outdated and current IPU configs are nested inside the same config files now), recommended to add newer IPU based tutorials


zhiyil-graphcore:
Hey thanks for the changes, I've tested this and I'm afraid pip install lightning gives lightning version of 2.0.4 which is no consistent with the 2.1.0 version that we have made changes with. The toymix config are not running as expcted and gives this error:
`poptorch.poptorch_core.Error: In poptorch/source/dispatch_tracer/RegisterAtenOverloads.cpp:100: 'poptorch_cpp_error': There is no active dispatch`

zhiyil-graphcore:
This works ok for me:
put `lightning @ git+https://github.com/Lightning-AI/lightning@ca30fd7752582201a3966806c92e3acbbaf2a045`
in a separate file named lightning.txt and run that separately. @shenyangHuang

DomInvivo:
Let's ignore the unittest failing for now, it's the stanford datacenter that is down #365

DomInvivo:
@shenyangHuang before merging, can you undo the deletion of the DGL-based tutorials? It's gonna be taken care in issue #374

shenyangHuang:
- added `lightning.txt` and tested that the lightning is now installed to be `2.1.0.dev0`
- added back the dgl tutorials, to be updated in #374

PR #378, opened 2023-07-02
GPU Optimizations

luis-mueller:
## Changelogs

- Introduces `metrics_every_n_steps` to the predictor, to control the number of steps after which logging happens. This can improve training speed, depending on the hardware. The default is `1`, which is also the previous behavior.
- Adds `BCEWithLogitsLossIPU`, which is required for mixed-precision when using a BCE loss. Using this loss with `bce_logits_ipu` requires to set `last_activation: null` of the corresponding task head. In addition, this loss is also considered more numerically stable ([see here](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html))
- Adds small optimizations to the code, following [Fast performance tips by Lightning](https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html).
- Adds fp32 conversions to ops that are not supported with fp16 to avoid runtime errors when using mixed-precision training
- Adds two new config files `config_small_gcn_gpu.yaml` and `config_large_gcn_gpu.yaml` with config overrides that can be used as starter configs when running ToyMix and LargeMix on GPU, respectively.

These changes were tested on a single A100 GPU.

---

_Checklist:_

---


expts/main_run_multitask.py
```diff
@@ -29,21 +29,23 @@
 # WandB
 import wandb
 
+import torch; torch.set_float32_matmul_precision("medium")
```
luis-mueller, line 32: Optimization for e.g., A100s with tensor cores, doesn't require to lower precision to 16 in general.

graphium/ipu/ipu_losses.py
```diff
@@ -1,12 +1,12 @@
...
 from torch import Tensor
-from torch.nn import BCELoss, MSELoss, L1Loss
+from torch.nn import BCELoss, BCEWithLogitsLoss, MSELoss, L1Loss
 from torch._C import _infer_size
 from loguru import logger
 from graphium.trainer.losses import HybridCELoss
 
 
-class BCELossIPU(BCELoss):
+class BCELossIPU(BCEWithLogitsLoss):
```
luis-mueller, line 9: Change required for mixed-precision training.

graphium/nn/pyg_layers/pooling_pyg.py
```diff
@@ -33,7 +33,7 @@ def scatter_logsum_pool(x: Tensor, batch: LongTensor, dim: int = 0, dim_size: Op
     Returns:
         the pooled features tensor
     """
-    dim_size = int(batch.max().item() + 1) if dim_size is None else dim_size
+    dim_size = int(batch.max().detach() + 1) if dim_size is None else dim_size
```
luis-mueller, line 36: See: https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html#item-numpy-cpu

graphium/nn/pyg_layers/pooling_pyg.py
```diff
@@ -65,7 +65,7 @@ def scatter_std_pool(x: Tensor, batch: LongTensor, dim: int = 0, dim_size: Optio
...
```
luis-mueller, line 68: See: See: https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html#item-numpy-cpu

graphium/trainer/losses.py
```diff
@@ -64,7 +64,7 @@ def forward(self, input: Tensor, target: Tensor) -> Tensor:
         # we apply softmax on the raw logits first
         softmax_input = self.softmax(input)
         # [batch_size, n_classes] * [n_classes] ([0, 1, 2...n_brakets-1]) -> [batch_size]
-        regression_input = torch.inner(softmax_input, self.brackets.to(input.device))
+        regression_input = torch.inner(softmax_input.float(), self.brackets.to(input.device))
```
luis-mueller, line 67: Required for mixed-precision (otherwise dtypes do not match in `torch.inner`.

graphium/trainer/predictor.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         metrics_on_training_set: Optional[Dict[str, List[str]]] = None,
         flag_kwargs: Dict[str, Any] = None,
         task_norms: Optional[Dict[Callable, Any]] = None,
+        metrics_every_n_steps: Optional[bool] = 1,
```
luis-mueller, line 42: Especially for smaller models, computing stats and logging is huge bottleneck. This parameter lets the user trade speed for detailed stats at each batch.

graphium/trainer/predictor.py
```diff
@@ -495,7 +504,7 @@ def get_gradient_norm(self):
         for p in self.parameters():
             if p.grad is not None:
                 param_norm = p.grad.detach().data.norm(2)
-                total_norm += param_norm.item() ** 2
+                total_norm += param_norm.detach().cpu() ** 2
```
luis-mueller, line 507: See: https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html#item-numpy-cpu

graphium/trainer/predictor.py
```diff
@@ -608,7 +617,7 @@ def on_train_start(self):
 
     def get_progress_bar_dict(self) -> Dict[str, float]:
         prog_dict = {}
-        prog_dict["loss"] = self.task_epoch_summary.weighted_loss.item()
+        prog_dict["loss"] = self.task_epoch_summary.weighted_loss.detach().cpu()
```
luis-mueller, line 620: See: https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html#item-numpy-cpu

graphium/utils/tensor.py
```diff
@@ -179,7 +179,7 @@ def nan_median(input: Tensor, **kwargs) -> Tensor:
                 median = median.squeeze(d)
     else:
         if dim is None:
-            median = torch.nanmedian(input.flatten())
+            median = torch.nanmedian(input.flatten().float())
```
luis-mueller, line 182: `torch.nanmedian` not implemented for `float16`

graphium/utils/tensor.py
```diff
@@ -251,7 +251,7 @@ def nan_std(input: Tensor, unbiased: bool = True, **kwargs) -> Tensor:
         output: The resulting standard deviation of the tensor
     """
 
-    return torch.sqrt(nan_var(input=input, unbiased=unbiased, **kwargs))
+    return torch.sqrt(nan_var(input=input, unbiased=unbiased, **kwargs).float())
```
luis-mueller, line 254: `torch.sqrt` not implemented for `float16`

graphium/ipu/ipu_losses.py
DomInvivo: You have to be careful here. First, calling the class `BCELossWithLogitsIPU` for consistency. Then, changing the `spaces.py` file to include both `bce_with_logits` and `bce_with_logits_ipu`, and changing the YAML config accordingly. Also changing the output layer in the yaml `last_activation` parameter to `null`.

expts/main_run_multitask.py
DomInvivo: Is this a GPU specific command? Does it work when running on CPU? Is this the best place to have it, or perhaps in the `get_accelerator` if we find a GPU with mixed precision?

expts/neurips2023_configs/config_large_gcn_luis.yaml
```diff
@@ -0,0 +1,412 @@
...
+    datamodule:
+      args:
+        ipu_dataloader_training_opts:
+          mode: async
+          max_num_nodes_per_graph: 60 # train max nodes: 20, max_edges: 54
+          max_num_edges_per_graph: 100
+        ipu_dataloader_inference_opts:
+          mode: async
+          max_num_nodes_per_graph: 60 # valid max nodes: 51, max_edges: 118
+          max_num_edges_per_graph: 100
```
DomInvivo, line 20: Remove/ comment these lines

expts/neurips2023_configs/config_small_gcn_luis.yaml
```diff
@@ -0,0 +1,346 @@
...
+    datamodule:
+      args:
+        ipu_dataloader_training_opts:
+          mode: async
+          max_num_nodes_per_graph: 44 # train max nodes: 20, max_edges: 54
+          max_num_edges_per_graph: 80
+        ipu_dataloader_inference_opts:
+          mode: async
+          max_num_nodes_per_graph: 44 # valid max nodes: 51, max_edges: 118
+          max_num_edges_per_graph: 80
```
DomInvivo, line 20: same

graphium/data/multilevel_utils.py
```diff
@@ -22,6 +22,11 @@ def unpack(graph_data):
         elif isinstance(graph_data, list):
             return np.array(graph_data)
         elif isinstance(graph_data, np.ndarray):
+            if len(graph_data.shape) == 0:
+                graph_data = np.expand_dims(graph_data, 0)
+            if graph_data.shape[0] == 0:
+                graph_data = np.array([np.nan])
```
DomInvivo, line 28: This seems like a risky behaviour that could lead to a shape mismatch. In what cases is `graph_data` empty?

graphium/trainer/losses.py
DomInvivo: This could break things. you could perhaps do this?
```suggestion
        regression_input = torch.inner(softmax_input.to(self.brackets.dtype), self.brackets.to(input.device))
```

graphium/utils/tensor.py
DomInvivo: This will change the dtype of the output. I would suggest casting back the `median` to its original dtype after the call to `.float()`

graphium/utils/tensor.py
DomInvivo: Same as above. Re-casting to the original dtype.

DomInvivo (changes requested):

expts/main_run_multitask.py
```diff
@@ -54,7 +56,7 @@ def main(cfg: dict, run_name: str = "main", add_date_time: bool = True) -> None:
     if add_date_time:
         date_time_suffix = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
 
-    wandb.init(entity="multitask-gnn", project=cfg["constants"]["name"], config=cfg)
+    wandb.init(entity=cfg["constants"]["entity"], project=cfg["constants"]["name"], config=cfg)
```
zhiyil-graphcore, line 60: I think here if would be good to have a default entity value if no "entity" is set in the config, otherwise it will complain with an `KeyError`. Most of our configs actually don't have the entity.

expts/main_run_multitask.py
```diff
@@ -115,4 +119,9 @@ def main(cfg: dict, run_name: str = "main", add_date_time: bool = True) -> None:
         CONFIG_FILE = args.config
     cfg = load_yaml_config(CONFIG_FILE, MAIN_DIR, unknown_args)
 
-    main(cfg)
+    for num_workers in [4, 8, 16, 24]:
+        for batch_size in [128, 256, 384, 512]:
```
zhiyil-graphcore, line 123: What's the intuition for this bit? Is it for GPU optimization?

graphium/trainer/predictor.py
```diff
@@ -161,6 +163,10 @@ def __init__(
         self.validation_step_outputs = []
         self.test_step_outputs = []
         self.epoch_start_time = None
+        
+        # Decide whether to log every step or once at the end 
+        # of the epoch.
+        self.metrics_every_n_steps = metrics_every_n_steps
```
zhiyil-graphcore, line 169: Is this related to the reporting of metrics during training? Reporting metrics every step might have some RAM issue? I wonder if we could set the default to be false?

graphium/trainer/predictor.py
zhiyil-graphcore, line 620: Not sure if it is related to the detach thing, I've tried to run this branch on IPU with the debug config, but it gives me strange compilation error. Is it possible to have .detach() (also other places) for CPU and GPU only?

zhiyil-graphcore (commented):
Thanks for the changes. I think some of the changes are not hardware agnostic, it would be good to keep these only for CPU and GPU.

luis-mueller:
Thanks @DomInvivo and @zhiyil-graphcore for you comments and suggestions. This is WIP, so there are still some artifacts from me experimenting in the code. I will address everything in the coming days, after I have gained more insight into the current bottlenecks on GPU.

@zhiyil-graphcore I agree with you that we should enable these changes exclusively on GPU (and perhaps CPU), although the `metrics_every_n_steps` parameter (computes and logs metrics only every n batches instead of at each batch) might also be speeding things up on IPU, right? (It is optional with n=1 being the default, so it should at least not break things)

zhiyil-graphcore:
> 

@luis-mueller Thanks for the reply. Yeah setting `metrics_every_n_steps` to be a larger value would definitely help with the RAM and speed. I think I confused this parameter with another one which is called `metrics_on_training_set`. So feel free to ignore my comments about `metrics_every_n_steps` previously.

expts/main_run_multitask.py
luis-mueller: I definitely agree. Had it there mostly to test it out. Moved it now as an optional feature to `load_accelerator` in the next commit.

graphium/data/multilevel_utils.py
luis-mueller: I observed this when parsing the PCQM4M data and added the missing case. A shape mismatch shouldn't occur because later in the same function we do nan-padding to fit the shapes again. If you still think this could be an issue, please elaborate on why it could lead to a shape mismatch. Perhaps I don't fully understand what you mean.

expts/main_run_multitask.py
luis-mueller: It was for testing purposes only. Will be removed in the next commit.

graphium/trainer/predictor.py
luis-mueller: I see. It is supposed to be a performance optimization, following [https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html#item-numpy-cpu](https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html#item-numpy-cpu). However, I'm not sure it improves a lot. If it causes cross-hardware issues, I'd rather undo it for now. Thanks for pointing this out.

expts/neurips2023_configs/config_large_gcn_luis.yaml
DomInvivo, line 1: These are ipu's configs for a GPU file.

Also, if you look at some recent config files, there is one config `config_large_gcn.yaml` that contains everything. The others, such as `config_large_gin.yaml` use the "override" parameter to change what's needed. Can you do that as well to avoid carrying so many large config files?

expts/neurips2023_configs/config_small_gcn_luis.yaml
DomInvivo, line 1: Same comment as before with the "override" parameter

graphium/config/_loader.py
```diff
@@ -455,6 +455,11 @@ def load_accelerator(config: Union[omegaconf.DictConfig, Dict[str, Any]]) -> Tup
     merge_dicts(config, config_override)
     accelerator_type = get_accelerator(config_acc)
 
+    if accelerator_type == "gpu":
+        precision = config_acc.get("float32_matmul_precision", None)
+        if precision is not None:
```
DomInvivo, line 460: Why? I don't understand the logic? From my perspective you can always do the `set_float32_matmul_precision` since if the user uses float16 or float64, it will be ignored anyways

graphium/data/multilevel_utils.py
DomInvivo: Is `graph_data.shape` giving you the number of nodes? 

If so, I know that there's a bug in our custom PCQM4M dataset where some molecules have a mismatched number of hydrogens, because hydrogens where always removed from the labels, but some molecules require explicit hydrogens (H2 for instance), which ended having 0 atoms.

Do you think it is related to that? If so, it's not the appropriate fix, but we could accept it temporarily.

graphium/nn/pyg_layers/pooling_pyg.py
DomInvivo, line 36: @zhiyil-graphcore can you confirm that the `detach` here won't cause issues on IPUs?

graphium/trainer/predictor.py
DomInvivo, line 507: @zhiyil-graphcore Can you confirm again that this won't cause issues on IPUs?

graphium/trainer/predictor.py
DomInvivo: @zhiyil-graphcore any update on that?

DomInvivo (changes requested):
Thanks @luis-mueller ! A few comments and changes, but mostly looks good.

Let's wait to see if the `detach` causes issues on IPU's and how to proceed.

expts/main_run_multitask.py
```diff
@@ -54,7 +55,7 @@ def main(cfg: dict, run_name: str = "main", add_date_time: bool = True) -> None:
     if add_date_time:
         date_time_suffix = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
 
-    wandb.init(entity="multitask-gnn", project=cfg["constants"]["name"], config=cfg)
+    wandb.init(entity="lmueller", project=cfg["constants"]["name"], config=cfg)
```
zhiyil-graphcore, line 58: I assume this entity name is just for debugging purpose?

graphium/data/datamodule.py
```diff
@@ -1252,9 +1252,17 @@ def _make_multitask_dataset(
...
         if stage == "train":
-            self.get_label_statistics(
-                self.processed_graph_data_path, self.data_hash, multitask_dataset, train=True
-            )
+            if self.processed_graph_data_path is not None:
+                data_path = self.processed_graph_data_path
+            elif self.cache_data_path is not None:
+                data_path = self.cache_data_path
+            else:
+                data_path = None
```
zhiyil-graphcore, line 1260: What's the difference between processed_graph_data_path and cache_data_path here?

graphium/nn/pyg_layers/pooling_pyg.py
zhiyil-graphcore: I assume pooling_pyg layer is being used by almost all the models? Just tested with toymix and largemix sample dataset, gcn model, it works fine on IPU.

graphium/trainer/predictor.py
zhiyil-graphcore: Is this reporting gradient norm? I looks like it is always reporting norm as 0 on IPU and the two debug cases with toymix and largemix are running ok on IPU.

graphium/trainer/predictor.py
zhiyil-graphcore: The compilation looks ok with this branch now. The compilation error might come from something else before.

graphium/data/datamodule.py
luis-mueller: When `cache_data_path` is set, the entire dataset is loaded into memory. If `processed_graph_data_path` is set, each graph is read from disk. In the particular function you marked, previously only the `processed_graph_data_path` was passed to obtain the label statistics. However, if one were to use `cache_data_path` (which is feasible for smaller datasets such as ToyMix), then without the above change, the label statistics are always re-calculated.

DomInvivo: As discussed today, let's remove `cache_data_path`, and always use `processed_graph_data_path`. When the reading is from RAM, simply pre-load everything into memory before training.

Also, don't forget to delete the old cache.

graphium/config/_loader.py
luis-mueller: My understanding of this line is that you can manually trade off between matmul-internal precision and speed (on some hardware). IMO this setting is user-specific, I wouldn't set it by default.
Source: https://pytorch.org/docs/stable/generated/torch.set_float32_matmul_precision.html#torch.set_float32_matmul_precision.

graphium/data/datamodule.py
luis-mueller: > Also, don't forget to delete the old cache.

I'm not sure I understand what you mean here.

graphium/data/multilevel_utils.py
luis-mueller: To be precise, `graph_data.shape` gives you the number of labels. So yes, when pre-processing the PCQM4M data, I encountered cases where no labels were given for the specific graph.

@zhiyil-graphcore Did you encounter any issues when pre-processing the LargeMix? Because if not, I'm confused why this caused issues on my side.

I can try to remove this again and see whether the error disappears with the newest changes in the main branch.

graphium/ipu/ipu_losses.py
```diff
@@ -1,11 +1,56 @@
 import torch
 from torch import Tensor
-from torch.nn import BCELoss, MSELoss, L1Loss
+from torch.nn import BCELoss, BCEWithLogitsLoss, MSELoss, L1Loss
 from torch._C import _infer_size
 from loguru import logger
 from graphium.trainer.losses import HybridCELoss
 
 
+class BCEWithLogitsLossIPU(BCEWithLogitsLoss):
```
luis-mueller, line 9: @DomInvivo Are you happy with this code duplication? I could try to refactor it.

graphium/trainer/predictor.py
luis-mueller: @zhiyil-graphcore Any updates here? Is this causing 0 norms being logged when running on IPU?
If so, let us switch back to `item()`, I don't think this impacts performance too much.

graphium/ipu/ipu_losses.py
```diff
@@ -1,11 +1,56 @@
...
+            new_size = _infer_size(target.size(), weight.size())
+            weight = weight.expand(new_size).clone()
+        else:
+            weight = torch.ones(target.shape, dtype=input.dtype, device=input.device)
+
+        # Replace the nan-targets by 0 or 1. Take the value closest to the input.
+        # Give a weight of 0 where there are nan-targets
+        nan_targets = target.isnan()
+        nan_targets_0 = (input < 0.5) & nan_targets
+        nan_targets_1 = (input >= 0.5) & nan_targets
```
zhiyil-graphcore, line 34: If there are raw logits as input, is 0.5 still the correct threshold?

graphium/config/_loader.py
```diff
@@ -455,6 +455,14 @@ def load_accelerator(config: Union[omegaconf.DictConfig, Dict[str, Any]]) -> Tup
...
+            torch.set_float32_matmul_precision(precision)
+
+    if accelerator_type != "ipu":
+        del config["predictor"]["optim_kwargs"]["loss_scaling"]
```
luis-mueller, line 464: The config override doesn't work for deleting keys (and I didn't find a straightforward fix). `loss_scaling` seems to be a argument specific to Graphcore optimizers

DomInvivo: Ahhh I see. That's a problem. I was thinking of creating a `config_update` and `config_override` that would have different behaviours, but then it makes things very complicated.

I don't like your current solution. But I guess we can proceed with the PR and open a new issue #404  (Nice issue number 😄 )

DomInvivo (approved):
I'm approving and will merge, but I have added a new issue #404

graphium/trainer/predictor.py
callumm-graphcore: I'd be surprised if this worked at all on IPU. How important is it?

luis-mueller: IMO it's not important. I simply implemented this as one of the changes suggested by [fast performance tips of lightning](https://pytorch-lightning.readthedocs.io/en/1.3.8/benchmarking/performance.html). If it causes IPU issues we should revert it back to `item()`.

PR #389, opened 2023-07-07
Update of CI workflow / coverage

s-maddrellmander:
## Changelogs

Combination of changes for the general CI improvements 
Going to cover:

- [ ] Coverage badges
- [ ] Tag the tests with different accelerators
- [ ] Additional test workflows for each type of accelerator 
- [ ] tbc.

---

_Checklist:_

---


hadim:
Hello @s-maddrellmander. I pushed https://github.com/datamol-io/graphium/commit/a6632dc2045c90d4456155693fd28c0fa4f192a6 on `main` to enable code coverage (will add the badge soonish). I prefer to keep the same structure as we have in our other repos.

As for the multiple CI (cpu, gpu, ipu), I am not sure if I understand the rationale behind those? The CI only runs on CPU on GitHub.

hadim:
Code cov is setup and I also pushed the badge (no need to commit in this repo). You can browse the coverage at https://app.codecov.io/gh/datamol-io/graphium

s-maddrellmander:
@hadim - that's great RE codecov, I really like the tool, so I'll leave that to you. Thanks!

> As for the multiple CI (cpu, gpu, ipu), I am not sure if I understand the rationale behind those? The CI only runs on CPU on GitHub.

This is the more important part, we want to be able to tag tests so we know what's working / what's not atm. 
There are a number of paths developing through the code (particularly CPU/GPU vs IPU) where data loading, loss functions, etc have some differences. What I think would be pretty good is if we had a main "tests" check that gives the base coverage. 
If that's passing then everything is passing. 
But it seems likely that there might be a case with changes causing the IPU tests to fail but GPU/CPU to be okay, and bing able to log and track that is important. 

Obviously atm there aren't too many specific cases where the tests as we currently have them would actually tell us this, but part of the plan for the next few weeks is I think to increase the coverage + add enough functionality to test IPU specific parts with the IPU model. My main thought here was just trying to get a feel for how this might look in the UX before committing one way or another. 

Would be very happy to have a chat with you about this next week!

hadim:
Thanks for the explanation but I still don't understand why you need to have multiple CIs for this.

If it's really mandatory then I am fine with a `test_ipu.yml` CI but please keep a single one for CPU and GPU `test.yml` for the moment. All those files need to be maintained and also complexity the infrastructure (as well as the CI time). So it's important to keep that to a bare minimum.

pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+markers =
```
hadim, line 2: Let's not create a new infra config file please. Move it to `pyproject.toml`:

```
[tool.pytest.ini_options]
markers = [
    "e2e: e2e tests that require a connection with GitLab",
    "integration: integration tests that have mocked external dependencies"
]
```

s-maddrellmander:
> Thanks for the explanation but I still don't understand why you need to have multiple CIs for this.

@hadim - it would be really great to talk to you offline / on slack about this. You're absolutely right about not wanting to waste compute / time, and I'd like to work out the best way to make it easiest for everyone to collaborate + keep the reliability of the library transparent.
If you want to send me a DM on slack my email is `samuelm@graphcore.ai`?

hadim:
@s-maddrellmander yes, you can find me on the M2D2 slack at "Hadrien Mary".

.github/workflows/test.yml
```diff
@@ -51,6 +52,28 @@ jobs:
       - name: Run tests
         run: pytest
 
+      - name: Test CLI
+        run: graphium --help
+
+      - name: Test building the doc
+        run: mkdocs build
+  
+  codecov:
```
hadim, line 61: Any reason for having a new section here instead of having just a coverage step above?

hadim: It seems best to have everything above especially cuz the test matrix would have to be defined twice.

.github/workflows/test.yml
```diff
@@ -51,6 +52,28 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10"]
+        pytorch-version: ["1.13"]
```
hadim, line 66: If graphium is compatible with pytorch 2 then I propose to test for pytorch 2 here (optionally we can test for both 1.13 and 2.

@DomInvivo any opinion?

.github/workflows/test_ipu.yml
```diff
@@ -0,0 +1,80 @@
...
+jobs:
+  test-ipu:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8"]
+        pytorch-version: ["2.0"]
+
+    runs-on: "ubuntu-latest"
+    timeout-minutes: 300
```
hadim, line 23: Possible to reduce the timeout? (a large upper bound is ok but I guess 300 mins is a bit too much)

.github/workflows/test_ipu.yml
```diff
@@ -0,0 +1,80 @@
...
+          python -m pip install --no-deps -e .
+
+      - name: Run tests
+        run: python3 -m pytest -m 'not skip_ipu'
+
+      - name: Test CLI
+        run: graphium --help
+
+  codecov:
+    strategy:
```
hadim, line 57: Same comment about having everything in the above section.

codecov.yml
```diff
@@ -5,3 +5,19 @@ coverage:
...
+  default_rules: # default rules that will be inherited by all components
+    statuses:
+      - type: project # in this case every component that doens't have a status defined will have a project type one
+        target: auto
+        branches:
+          - "!main"
+  individual_components:
+    - component_id: ipu # this is an identifier that should not be changed
+      name: ipu # this is a display name, and can be changed freely
+      paths:
```
hadim, line 22: @s-maddrellmander I am just testing if we can get coverage separated from the rest. Will remove if it does not work.

pyproject.toml
```diff
@@ -102,6 +102,13 @@ filterwarnings = [
     "ignore::DeprecationWarning:pytorch_lightning.*:",
     "ignore::DeprecationWarning:pkg_resources.*:",
 ]
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+    "fast: marks tests as fast (select with '-m \"fast\"')",
+    "cpu: marks tests that are specific to the CPU (deselect with '-m \"not cpu\"')",
+    "gpu: marks tests that are specific to the GPU (deselect with '-m \"not gpu\"')",
+    "ipu: marks tests that are specific to the IPU (deselect with '-m \"not ipu\"')",
```
hadim, line 110: Should/can we remove the unused ones? cpu, gpu, fast, slow?

tests/test_dummy_gpu.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
```
hadim, line 2: can this file be removed?

tests/test_ipu_dataloader.py
```diff
@@ -177,127 +187,26 @@ def test_poptorch_graphium_deviceiterations_gradient_accumulation(self):
...
-        opts.deviceIterations(device_iterations)
-        training_opts = deepcopy(opts)
-        training_opts.Training.gradientAccumulation(gradient_accumulation)
-        inference_opts = deepcopy(opts)
-
-        # Load the configuration file for the model
-        CONFIG_FILE = "tests/config_test_ipu_dataloader.yaml"
+        # Simplified testing config - reflecting the toymix requirements
+        # CONFIG_FILE = "tests/config_test_ipu_dataloader.yaml"
+        CONFIG_FILE = "tests/config_test_ipu_dataloader_multitask.yaml"
```
hadim, line 192: Usually it's best to use the pytest fixture instead of absolute paths: https://github.com/datamol-io/graphium/blob/5397d66e31ec44a656e74ca7f3517f2e074d3679/tests/conftest.py#L10

This is minor, feel free to ignore.

hadim (changes requested):
Thanks @s-maddrellmander .

See a few comments below. The main things is the proposal to unify the test within a single section instead of having a separate one (unless I missed something that require this).

Besides that it sounds great.

codecov.yml
hadim: It does not seem to work. We can still keep it and merge with it. I will take care of it on `main` (it's possible you need this config to be on `main` for this to work).

.github/workflows/test.yml
s-maddrellmander: Nope, that's true. I was trying to make it possible to use the full coverage based on the two env tests, but I think this is a simpler solution.

.github/workflows/test.yml
s-maddrellmander: I've updated this to 2.0 here as well.

hadim:
@s-maddrellmander dont worry about pytorch>=2, I will take care of it later.

hadim (approved):
Thank you for your patience!

hadim:
codecov components seem to work:

image

PR #391, opened 2023-07-11
Largemix convergence, throughput optimisation and add sampler

zhiyil-graphcore:
## Changelogs

---

_Checklist:_

---

- [x] Added a DatasetSubSampler to sample a percentage of the large dataset and combine with the small dataset for multi-task training.
- [x] Modified config to use fp16 and fit large micro batch size as well as larger device iteration to boost the throughput.
- [x] Rolled in some changes from @joao-alex-cunha and @callumm-graphcore about removing preds and targets during training to further boost the throughput. 
- [x] Added a BCEWithLogitsLossIPU which resolved the nan issue in pcba.
- [x] Fixed a bug in HybridCELoss about applying softmax on 0s, which could lead to large losses for vcap and mcf7.
- [x] 0 hybriceloss on certain replica is fixed by filtering 0s before averaging 
- [x] inference is set to be using 1 replica, 1 device iter and gradient accumulation 1, so the inference metrics and losses can be fully trusted. 
- [x] Summary: 
1, with the current optimization, we could reach 7x speedup for largmix full dataset and it takes **10 hours for 50 epochs**. 
2, **mcf7 and vcap are training** with the fix in hybridceloss
Screenshot 2023-07-14 at 12 43 07


graphium/data/datamodule.py
```diff
@@ -766,6 +767,7 @@ def __init__(
         featurization_progress: bool = False,
         featurization_backend: str = "loky",
         featurization_batch_size: int = 1000,
+        sampler_task_dict: Optional[dict] = None,
```
DomInvivo, line 770: The value of `sampler_task_dict` is set to `Optional`, but you don't handle the case where it is `None`.

Also, as mentioned above, it's better if it's in the `task_specific_args`.
This means that the ratios will be assigned to the `SingleTaskDataset`, and then you can create the `sampler_task_dict` from the `MultiTaskDataset`. If a value is missing, the correct default should be `1.0`, where you take the full dataset.

graphium/data/sampler.py
```diff
@@ -0,0 +1,27 @@
...
+        indices = []
+        for task_name in self.sampler_task_dict.keys():
+            task_size = int(len(self.task_indices[task_name]) * self.sampler_task_dict[task_name])
+            indices += np.random.choice(self.task_indices[task_name], task_size, replace=False).tolist()
+        return iter(indices)
+
+    def __len__(self):
+        total_size = 0
+        for task_name in self.sampler_task_dict.keys():
+            total_size += int(len(self.task_indices[task_name]) * self.sampler_task_dict[task_name])
```
DomInvivo, line 26: This is wrong considering the comment above. If molecules intersect between sets, the len here would be larger than the len of the dataset.

graphium/data/sampler.py
```diff
@@ -0,0 +1,27 @@
...
+        self.dataset = dataset
+        self.sampler_task_dict = sampler_task_dict
+        self.task_indices = {}
+        for i in range(len(dataset)):
+            task_name = dataset[i]["label"].name()
+            if task_name not in self.task_indices:
+                self.task_indices[task_name] = []
+            self.task_indices[task_name].append(i)
+
+    def __iter__(self):
```
DomInvivo, line 16: Is `__iter__` reinitialized every epoch or only at the beginning? Because we want the random sampling to change every epoch.

DomInvivo (changes requested):
Few mistakes and changes with the `CustomSampler`

graphium/data/sampler.py
zhiyil-graphcore: I think it is reinitialized for each epoch. According to this discussion [here](https://stackoverflow.com/questions/75215951/pytorch-dataloader-across-epoch), the sequencial sampler and random sampler is applied for each epoch. And according to the pytorch [source code](https://pytorch.org/docs/stable/_modules/torch/utils/data/sampler.html#Sampler), the `CustomSampler` inherited from `data_utils.Sampler`, which has the `__iter__`, similar as the `SequentialSampler` and `RandomSampler`.

expts/neurips2023_configs/debug/config_large_gcn_debug.yaml
```diff
@@ -155,6 +155,7 @@ datamodule:
     num_workers: 30 # -1 to use all
     persistent_workers: False # if use persistent worker at the start of each epoch.
     # Using persistent_workers false might make the start of each epoch very long.
+    sampler_task_dict: {"graph_l1000_vcap": 1.0, "graph_l1000_mcf7": 1.0, "graph_pcba_1328": 0.5, "graph_pcqm4m_g25": 0.5, "node_pcqm4m_n4": 0.5}
```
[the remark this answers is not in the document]
zhiyil-graphcore: Yeah good idea, I'll move it to be under the `atamodule.task_specific_args.[DATASET_NAME]` section.

graphium/data/sampler.py
```diff
@@ -0,0 +1,27 @@
...
+        indices = []
+        for task_name in self.sampler_task_dict.keys():
+            task_size = int(len(self.task_indices[task_name]) * self.sampler_task_dict[task_name])
+            indices += np.random.choice(self.task_indices[task_name], task_size, replace=False).tolist()
+        return iter(indices)
```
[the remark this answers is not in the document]
zhiyil-graphcore: Oh I see, does this point to the pcqm dataset for the `g25` and `n4` tasks?  That'a a very good point!

graphium/data/sampler.py
zhiyil-graphcore: I've made some changes to it, in which we get the `total_size` from `__iter__` and use it in `__len__`, that seems to run ok when I'm testing with the debug config.

graphium/data/sampler.py
```diff
@@ -0,0 +1,27 @@
...
+import numpy as np
+
+
+class CustomSampler(data_utils.Sampler):
+    def __init__(self, dataset, sampler_task_dict):
+        self.dataset = dataset
+        self.sampler_task_dict = sampler_task_dict
+        self.task_indices = {}
+        for i in range(len(dataset)):
+            # the dataset[i]["labels"].keys() are not deterministic, need to sort by key length
```
zhiyil-graphcore, line 11: I suspect this for loop would take too long, even though it does only once at the beginning. Thinking of a way to cache it and reload.

zhiyil-graphcore: cache added, now use much less time with the `__init__` part.

graphium/data/sampler.py
```diff
@@ -0,0 +1,43 @@
+import torch.utils.data as data_utils
+import numpy as np
+import time
+import os
+import torch
+
+
+class CustomSampler(data_utils.Sampler):
```
DomInvivo, line 8: I suggest changing the name from `CustomSampler` to `DatasetSubSampler` or something similar that tells you more about the class

graphium/data/sampler.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, dataset, sampler_task_dict, data_path, data_hash):
```
DomInvivo, line 9: Please add the docstring here to explain what the class does.

Also, define the type of each of the inputs. Something like that
```dataset: Dataset, sampler_task_dict: Dict[str, Optional[float]], data_path: PathLike, data_hash: str```

File: graphium/data/sampler.py [hunk 9]
```diff
@@ -0,0 +1,43 @@
...
+            elapsed = round(time.time() - now)
+            print(f"Sampler-->time spent on getting indices: {elapsed}.")
+            torch.save(self.task_indices, filename, pickle_protocol=4)
+
+    def __iter__(self):
+        indices = []
+        for task_name in self.task_indices.keys():
+            task_size = int(len(self.task_indices[task_name]) * self.sampler_task_dict[task_name])
+            indices += np.random.choice(self.task_indices[task_name], task_size, replace=False).tolist()
+            indices_set = set(indices)
```
DomInvivo, line 38: Minor suggestion to make the code more readable
```suggestion
            indices += np.random.choice(self.task_indices[task_name], task_size, replace=False)
            indices_set = set(indices.tolist())
```

DomInvivo (changes requested):

graphium/data/datamodule.py
zhiyil-graphcore: Cool, I've made the changes accordingly. The only thing I could think of is that this approach would need the cache to be regenerated each time we change the `epoch_sampling_fraction`, even though the cache does not actually need to be regenerated. Let me know your thoughts.

graphium/trainer/losses.py
```diff
@@ -63,6 +63,8 @@ def forward(self, input: Tensor, target: Tensor) -> Tensor:
         # regression loss needs normalized logits to probability as input to do inner product with self.brackets
         # we apply softmax on the raw logits first
         softmax_input = self.softmax(input)
+        # the softmax of a tensor of 0s would not be 0s anymore, so need to apply nan_targets here again to filter out
+        softmax_input[nan_targets] = 0.0
```
zhiyil-graphcore, line 67: Found a bug here: the nans are set to 0s in `ipu_losses.py` and here `softmax` is applied, while the `softmax` of 0s are not 0s. Previous approach would lead to wrong estimation of hybridceloss.

graphium/data/sampler.py
```diff
@@ -0,0 +1,65 @@
...
+                self.task_indices[task_name].append(i)
+            elapsed = round(time.time() - now)
+            print(f"Sampler-->time spent on getting indices: {elapsed}.")
+            torch.save(self.task_indices, filename, pickle_protocol=4)
+
+    def __iter__(self):
+        indices = []
+        for task_name in self.task_indices.keys():
+            task_size = int(len(self.task_indices[task_name]) * self.sampler_task_dict[task_name])
+            indices += np.random.choice(self.task_indices[task_name], task_size, replace=False).tolist()
```
zhiyil-graphcore, line 51: Had to use the `.tolist()` here as it was numpy array, before `tolist` and not happy with `+=`.

graphium/trainer/losses.py
```diff
@@ -52,22 +52,40 @@ def __init__(
...
         Parameters:
             input: (batch_size x n_classes) tensor of logits predicted for each bracket.
             target: (batch_size) or (batch_size, 1) tensor of target brackets in {0, 1, ..., self.n_brackets}.
         """
 
         target = target.flatten()
+
+        # set input and target with nans to 0s for regression loss
+        input[nan_targets] = 0.0
+        target[nan_targets] = 0.0
```
callumm-graphcore: I was worried this wouldn't work on IPU due to dynamic shapes, but it does work!

graphium/trainer/losses.py
```diff
@@ -52,22 +52,42 @@ def __init__(
...
 
+            # set input and target with nans to -1000s for ce loss
+            input[nan_targets] = -1000
+            target[nan_targets] = -1000
         # cross_entropy loss needs raw logits as input
-        ce_loss = F.cross_entropy(input, target.long(), weight=self.weight, reduction=self.reduction)
+        # ce_loss does not need scaling as it already ignores -1000 masked nan values
+        ce_loss = F.cross_entropy(
+            input, target.long(), weight=self.weight, ignore_index=-1000, reduction=self.reduction
+        )
```
zhiyil-graphcore, line 91: Found a bug here: the nans are set to 0s in `ipu_losses.py` and here `softmax` is applied inside regression and ce losses, while the `softmax` of 0s are not 0s. Previous approach would lead to wrong estimation of hybridceloss. 
1, for regression_loss, it is easy to filter that out by nan_targets
2, for ce_loss softmax is integrated with the `F.cross_entropy`, so used -1000 to mark the nans and ask the ce loss to ignore -1000.
3, this means ce_loss does not need nan_target scaling any more, so we do scaling only on regression_loss

graphium/trainer/predictor.py
```diff
@@ -447,17 +447,17 @@ def on_train_batch_end(self, outputs, batch: Any, batch_idx: int) -> None:
...
+        #     step_name="train",
+        #     targets=outputs["targets"],
+        #     predictions=outputs["preds"],
+        #     loss=outputs["loss"],  # This is the weighted loss for now, but change to task-specific loss
+        #     task_losses=outputs["task_losses"],
+        #     n_epochs=self.current_epoch,
+        # )
+        # metrics_logs = self.task_epoch_summary.get_metrics_logs()  # Dict[task, metric_logs]
+        # metrics_logs["_global"]["grad_norm"] = self.get_gradient_norm()
+        # outputs.update(metrics_logs)  # Dict[task, metric_logs]. Concatenate them?
```
callumm-graphcore, line 460: Are you sure commenting this out is a good idea?

graphium/trainer/losses.py
```diff
@@ -52,22 +52,40 @@ def __init__(
         self.alpha = alpha
         self.softmax = torch.nn.Softmax(dim=1)
 
-    def forward(self, input: Tensor, target: Tensor) -> Tensor:
+    def forward(self, input: Tensor, target: Tensor, nan_targets: Tensor = None) -> Tensor:
```
callumm-graphcore, line 55: Do the other loss functions not take a `nan_targets` argument?

graphium/trainer/predictor.py
zhiyil-graphcore: Not sure, I'd like to discuss with @DomInvivo about how to make that change properly, I think it might be useful for something.

graphium/trainer/losses.py
zhiyil-graphcore: I think the other loss functions do not need `nan_targets` inside forward, it can be applied after in `ipu_losses.py`.

DomInvivo (changes requested):
Thanks @zhiyil-graphcore for the awesome PR! So happy to see the speed gains and the loss going down!! 😄 

I have a few comments, but I hope to merge it very soon

graphium/data/datamodule.py
DomInvivo: You can exclude the `epoch_sampling_fraction` from the hash key. I `get_data_hash` you can pop it from `task_specific_args`

DomInvivo:
@zhiyil-graphcore there are a few changes that happened on the main that can impact this PR. I suggest merging main in your branch to update it and resolve conflicts. Especially:

1. There's now a conflict for setting the `self.task_specific_args`, since the main now enforces the `DatasetProcessingParams` object to be created.
2. There's also the PR #378 that added the parameter `metrics_every_n_steps` that will allow you to uncomment the `on_train_batch_end`.

DomInvivo:
Also, @zhiyil-graphcore , is the 10h per 50 epochs considers the fact that epochs are now smaller with the `epoch_sampling_fraction`? Or is it using a value == 1?

zhiyil-graphcore:
> Also, @zhiyil-graphcore , is the 10h per 50 epochs considers the fact that epochs are now smaller with the `epoch_sampling_fraction`? Or is it using a value == 1?

That was with fulldataset, not using sampling at all. mcf7 and vcap are actually training without sampling and it takes 19 hours for 100 epochs, fp16, no nans. I think we have further potential to speedup if there is a way to improve the packing strategy, which Callum is working on.

graphium/ipu/ipu_utils.py
```diff
@@ -113,7 +113,7 @@ def load_ipu_options(
         ), f"Received inconsistent gradient accumulation `{current}` and `{gradient_accumulation}"
         ipu_options.Training.gradientAccumulation(gradient_accumulation)
 
-    ipu_options.anchorTensor("input", "input")
+    # ipu_options.anchorTensor("input", "input")
```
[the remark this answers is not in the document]
zhiyil-graphcore: Commenting that out prevents a strange error from happening on IPU during compilation and that error is not deterministic, it have a certain probability to appear. We were not fully sure what this line is for: `ipu_options.anchorTensor("input", "input")`, so leaving it there like that. If it doesn't mean anything to you we could just remove it. Not sure why it was added to the code.

graphium/ipu/ipu_losses.py
```diff
@@ -131,14 +176,7 @@ def forward(self, input: Tensor, target: Tensor) -> Tensor:
...
-        target[nan_targets] = 0.0
 
         # Compute the loss, and rescale by the number of nan elements
-        loss = super().forward(input, target)
-        num_real_targets = (~nan_targets).sum()
-        factor1 = torch.where(num_real_targets > 0, 1, 0)
-        factor2 = torch.where(num_real_targets > 0, 0, 1)
-        loss = factor1 * loss * nan_targets.numel() / (num_real_targets + factor2)
-
+        loss = super().forward(input, target, nan_targets)
```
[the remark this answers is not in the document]
zhiyil-graphcore: I think the `nan_targets` are needed inside the forward of the loss function, because first time nans are set to 0s, but after applying softmax, 0s will not stay as 0. This will make the regression loss very large. So we apply `nan_targets` again after softmax.

graphium/data/datamodule.py
```diff
@@ -690,6 +692,7 @@ def __init__(
         self.splits_path = splits_path
         self.split_names = split_names
         self.label_normalization = label_normalization
+        self.epoch_sampling_fraction = epoch_sampling_fraction
```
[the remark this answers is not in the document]
zhiyil-graphcore: Yeah, I agree. But it doesn't have to be `<=1`? It can be like 2.0 if we plan to upsample some dataset?

graphium/data/datamodule.py
```diff
@@ -657,6 +658,7 @@ def __init__(
         splits_path: Optional[Union[str, os.PathLike]] = None,
         split_names: Optional[List[str]] = ["train", "val", "test"],
         label_normalization: Optional[Union[Dict[str, Any], omegaconf.DictConfig]] = None,
+        epoch_sampling_fraction: Optional[float] = 1.0,
```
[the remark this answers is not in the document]
zhiyil-graphcore: I think `None` can be the same as 1.0, it makes the checking easier. Both are fine for me though. I can remove the `Optional` and make the default `epoch_sampling_fraction` to be 1.0.

expts/neurips2023_configs/config_large_gcn_sampler.yaml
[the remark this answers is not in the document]
zhiyil-graphcore: Yeah you're right after removing `epoch_sampling_fraction` from hash, we can just add `epoch_sampling_fraction` to the config file and set to 1.0 as default. Maybe we also need to let the user know that the number of epochs need to scale up if subsampling is used?

graphium/trainer/losses.py
zhiyil-graphcore: I think we need the `F.cross_entroy` as it has labels of [0, 1, 2, 3, 4]? Is `BCEwithLogitsLossIPU` just used for binary cross entropy? For regression loss, I think indeed we can pass the logits after softmax to the `MSELossIPU`. I think the point here is that, for regression loss, input[nan_target] = 0.0 and softmax_input[nan_target] = 0.0 need to be applied, for ce loss, `F.cross_entropy` includes a softmax, we can't set nans to 0s and hope that the ce loss would ignore that as softmax of 0s is not 0s.

graphium/data/datamodule.py
zhiyil-graphcore: good point!

graphium/ipu/ipu_losses.py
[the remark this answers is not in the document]
DomInvivo: Instead of setting to 0, can you set to -inf? Or to -1000?

graphium/ipu/ipu_utils.py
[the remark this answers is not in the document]
DomInvivo: I would delete it, if you're not sure what it means

DomInvivo:
> The only caveat left: the filtering of nans is on micro batch size, the global batch size has: gradient accumulation and replica. for combined batch size, there is also device iteration. this means for mcf7 and vcap, if there is no sample inside a certain micro batch size, it would be all nans, and the loss would be 0.

This was something we discussed since time ago but without a proper solution. Using `epoch_sampling_fraction` should help a little. 

Otherwise, we could have a counter during `training_step` that counts the number of rows without nans for each task, that can then be multiplied for the accumulation. It gets a bit more tricky for node/edge/nodepair level tasks.  The problem is that, on IPU, I think the code for accumulation is hidden I believe???

graphium/ipu/ipu_losses.py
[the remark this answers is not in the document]
zhiyil-graphcore: I think the issue is that unlike masking for the transformer module, we have [0.2, 1.5, 3.2, 0.3, -1000], applying softmax on -1000 would give us 0. In this case it is like [nan, nan, nan] --> [-1000, -10000, -10000] ----> softmax([-1000, -10000, -10000]) --> [0.333, 0.333, 0.333]. It ends up the same as 0s. `F.cross_entropy` takes the `ignore_index` to ignore index -1000. We could also use the `self.weight` parameter like the BCE loss. The l1_loss and mse_loss, on the other hand, does not take `ignore_index` or  `self.weight`.  Maybe we could create a CELossIPU and calculate hybridceloss as: alpha*CELossIPU + (1-alpha)*MSELossIPU

graphium/trainer/predictor.py
zhiyil-graphcore: I think this is slightly different, by using `metrics_every_n_steps`, like we set to 100, we only report the metrcis, including loss every 100 steps, this might not be helpful with the wandb reporting if we wanna track the progress of the training as we only have 157 steps in one epoch. While this change above is about if the "preds" and "targets" are saved each step. I think it would be good to combine the two:
Set `metrics_every_n_steps` to 3 or 5 + remove "preds" and "targets" from step_dict.

graphium/ipu/ipu_wrapper.py
```diff
@@ -101,7 +101,7 @@ def compute_loss(
 
     def on_train_batch_end(self, outputs, batch, batch_idx):
         outputs = self.convert_from_fp16(outputs)
-        outputs["loss"] = outputs["loss"].mean()
+        outputs["loss"] = outputs["loss"][outputs["loss"] != 0].mean()
```
DomInvivo, line 104: I'm really not sure about this change. This will cause an issue if the loss is actually 0 (which is very unlikely).

But what I'm most concerned about is that this only changes the loss that gets logged, not the gradient accumulation. 

Here's an example of how gradient accumulation is done in pytorch:

```python
optimizer.zero_grad()
for i, (inputs, labels) in enumerate(train_loader):
    outputs = model(inputs)
    loss = criterion(outputs, labels)
    loss = loss / accumulation_steps
    loss.backward()
    if (i + 1) % accumulation_steps == 0:
        optimizer.step()
        optimizer.zero_grad()
```

You can see that the loss gets divided by the `accumulation_steps` early on, but it's only the `backward` passes that gets accumulated later. If, after you're done with your steps, but before the call to `optimizer.step()`, you decide to rescale the loss, then it's too late.

DomInvivo: Instead, the better way of doing this is:
1. For all tasks, check if `labels.isnan().all()`, instead of checking if the loss is `0`
2. If `True` for `task_K`, then you can detach the task head `self.model.task_heads[task_K].requires_grad = False`

But because of the `if` statements, I don't think this is actually possible when compiled on IPU's.

When we discussed last time to check if 0-loss meant 0-gradient, I think the answer is yes since it means the model is perfect.

The issue then comes from the division by `accumulation_steps`.

DomInvivo (commented):
Wrong behaviour for the gradient descent, but right behaviour for the metric logging

graphium/ipu/ipu_wrapper.py
zhiyil-graphcore: The loss would not be actually 0 because we do gradient descent on summation of 5 task losses? The 0 loss would be mostly on the mcf7 and vcap I think?

Issue #360, opened 2023-06-16
Hydrogen in dataset error


zhiyil-graphcore:
When removing hydrogen, some molecules end up empty. in `datamodule.py `
`df = df[df[df.columns[0]].apply(lambda x: has_atoms_after_h_removal(x))]` was used temporarily, but this does not work for the test and PCQM4M dataset. need a proper fix.

DomInvivo:
In the `prepare_data` method, we filter the molecules that failed featurization here:
https://github.com/datamol-io/graphium/blob/ffe3ce6320ceab1ebc6a5cdee8cba2e2f9eadce5/graphium/data/datamodule.py#L1061

However, some molecules can succeed featurization, but when you then need to make sure that the number of nodes/edges in the graph match the number of nodes/edges expected in the labels, if not filter them out.

- Loop through all the molecules
  - Loop through all the labels that start with `node_`, `edge_` and `nodepair_` to get the expected number of nodes and edges
  - Ignore all the `NaN`s or `None`s in these label columns.
  - Loop through all the `features` which are supposed to be either a `GraphDict` or `pyg.data.Data`. And get the number of nodes and edges with the properties `.num_nodes` and `.num_edges`.
  - Make sure that all number of nodes and number of edges are consistent. 
  - Every time that you find an inconsistency, replace the given element in `features` by a string explaining the error `f"Inconsistent number of nodes found for '{this_smiles}': {set(all_num_nodes)}"

Issue #357, opened 2023-06-16
Update Docstrings


shenoynikhil:
No mention of this in docstrings, can docstrings be updated?
https://github.com/datamol-io/graphium/blob/2790af4d95e8924671cc666fac002fd003130618/graphium/data/dataset.py#L149

PR #400, opened 2023-07-13
Checking for inconsistent numbers of nodes/edges/nodepairs between labels and features

WenkelF:
@DomInvivo I implemented a function checking for inconsistent numbers of nodes/edges/nodepairs between labels and features in the datamodule for #360. To keep label information separate from the featurization process, I created a new function `found_size_mismatch` in addition to the existing `did_featurization_fail`.

Otherwise, it could be included [here](https://github.com/datamol-io/graphium/blob/ffe3ce6320ceab1ebc6a5cdee8cba2e2f9eadce5/graphium/data/datamodule.py#L1524-L1534)

Also fixed docstrings according to #357.


graphium/data/utils.py
```diff
@@ -105,3 +113,42 @@ def get_keys(pyg_data):
...
+    mismatch = False
+
+    if task.startswith("graph_"):
+        pass
+
+    elif task.startswith("node_"):
+        if labels.shape[0] != features.num_nodes:
+            mismatch = True
+            logger.warning(
+                (f"Inconsistent number of nodes between labels and features for {smiles}: {labels.shape[0]} vs {features.num_nodes}")
```
DomInvivo, line 133: Add the `task` to the warning message

DomInvivo (commented):
Looks good for me, but it fails the unit-tests for the `test_multitask_datamodule`. Can you check if some molecules get filtered out accidentally? Or perhaps the test needs updating?

WenkelF:
@DomInvivo the test for the unit-tests for the test_multitask_datamodule is fixed in f39585370753073f7e700bc2b790ef146d151235

The `_filter_none_molecules` function is now applied to the union of none molecules across tasks. Otherwise (filtering per task), causes the random train/val/test split to assign molecules to more than one (with different tasks), blowing up the size of all three, i.e., #train + #val + #test > #all.

It's a result of different `task_dataset_args[task]["sample_idx"]` per task below:
https://github.com/datamol-io/graphium/blob/ba974b210ff62706d17e979d212abbead613235f/graphium/data/datamodule.py#L1116-L1124

Could this cause an issue when combining different datasets with shared molecules?

graphium/data/datamodule.py
```diff
@@ -1075,12 +1075,16 @@ def has_atoms_after_h_removal(smiles):
...
             # Find out which molecule failed featurization, and filter them out
-            idx_none = []
-            for idx, feat in enumerate(args["features"]):
-                if did_featurization_fail(feat):
+            for idx, (feat, labels, smiles) in enumerate(zip(args["features"], args["labels"], args["smiles"])):
+                if did_featurization_fail(feat) or found_size_mismatch(task, feat, labels, smiles):
                     idx_none.append(idx)
+        idx_none = list(set(idx_none))
+
+        for task, args in task_dataset_args.items():
```
DomInvivo, line 1087: I don't think what you did makes sense. If I understand correctly, if the 4th molecule on task1 failed featurization, it will also remove the 4th molecule on task2. But these might be completely different molecules.

tests/test_multitask_datamodule.py
```diff
@@ -257,9 +257,9 @@ def test_multitask_with_missing_fromsmiles_from_config_parquet(self):
         dm.setup()
 
         # self.assertEqual(len(dm), 100)                      # Should this have a fixed value for when it's initialized? MTL dataset only gets created after.
-        self.assertEqual(len(dm.train_ds), 602)  # type: ignore
-        self.assertEqual(len(dm.val_ds), 201)  # type: ignore
-        self.assertEqual(len(dm.test_ds), 201)  # type: ignore
+        self.assertEqual(len(dm.train_ds), 596)  # type: ignore
+        self.assertEqual(len(dm.val_ds), 199)  # type: ignore
+        self.assertEqual(len(dm.test_ds), 199)  # type: ignore
```
DomInvivo, line 262: Why do you change the length here? Are there molecules that fail the `found_size_mismatch` part? Which ones? Can we see the warning?

DomInvivo (commented):
I'm not sure about the changes. Let's discuss that quickly on Monday

DomInvivo:
Let's not forget to remove the temporary fix described in #360 : https://github.com/datamol-io/graphium/issues/360#issue-1760806251

graphium/data/utils.py
WenkelF: The warning message has been updated here c7b68cd2810b7b6c0b9259f814e4f969b946232e

graphium/data/datamodule.py
WenkelF: You are right. After our discussion, it was not the filtering, but rather the unit test `tests/test_multitask_datamodule.py` that needed to be fixed here cde61e166572b94b8718f6dba5c3e39a03839503

WenkelF:
@DomInvivo this should be ready to merge

PR #407, opened 2023-07-17
improve inconsistent dimensions error messages

joao-alex-cunha:
I've stumbled on this error and found it very cryptic. The error message is the same for inconsistent node or edge dimensions which makes it worse. I've added references to `pre-NN` and `pre-NN-edges` and include the inconsistent values in the error message. Hopefully this makes it easier for the user to understand what is wrong in the config.


s-maddrellmander (approved):
LGTM!

PR #408, opened 2023-07-18
Using 16-true precision

joao-alex-cunha:
With Lightning 2.0, `precision: 16` means mixed precision (activations are computed in fp16 but weight update is done in fp32). For the IPU we want to train in true 16 precision, which Lightning represents as "16-true".

Also, precision needs to be passed to the function that sets IPUOptions, in order to set half-partials.

I've only corrected the precision in the debug configs. More configs should be fixed.

### DIFF
--- a/graphium/config/_loader.py
+++ b/graphium/config/_loader.py
@@ -123,7 +123,7 @@ def load_datamodule(
             model_name=config["constants"]["name"],
             gradient_accumulation=config["trainer"]["trainer"].get("accumulate_grad_batches", None),
             ipu_inference_opts=ipu_inference_opts,
-            precision=config["trainer"]["trainer"].get("precision")
+            precision=config["trainer"]["trainer"].get("precision"),
         )
         # Define the Dataloader options for the IPU on the training sets
         bz_train = cfg_data["batch_size_training"]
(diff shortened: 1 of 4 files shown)


graphium/config/_loader.py
```diff
@@ -123,6 +123,7 @@ def load_datamodule(
             model_name=config["constants"]["name"],
             gradient_accumulation=config["trainer"]["trainer"].get("accumulate_grad_batches", None),
             ipu_inference_opts=ipu_inference_opts,
+            precision=config["trainer"]["trainer"]["precision"]
```
DomInvivo, line 126: To handle when precision is not given

```suggestion
            precision=config["trainer"]["trainer"].get("precision")
```

graphium/config/_loader.py
```diff
@@ -366,6 +367,7 @@ def load_trainer(
             seed=config["constants"]["seed"],
             model_name=config["constants"]["name"],
             gradient_accumulation=config["trainer"]["trainer"].get("accumulate_grad_batches", None),
+            precision=config["trainer"]["trainer"]["precision"]
```
DomInvivo, line 370: To handle when precision is not given

```suggestion
            precision=config["trainer"]["trainer"].get("precision")
```

DomInvivo (approved):

PR #409, opened 2023-07-18
Changed the metrics logging logic

DomInvivo:
@zhiyil-graphcore and @luis-mueller , I merged your logics of how to reduce training time by not returning large but unused tensors. This is on top of PR #391

### DIFF
--- a/graphium/trainer/losses.py
+++ b/graphium/trainer/losses.py
@@ -63,14 +63,14 @@ class HybridCELoss(_WeightedLoss):
 
         # set input and target with nans to 0s for regression loss
         if nan_targets is not None:
-            input[nan_targets] = 0.0
-            target[nan_targets] = 0.0
+            input = torch.masked_fill(input, nan_targets.unsqueeze(1), 0)
+            target = torch.masked_fill(target, nan_targets, 0)
         # regression loss needs normalized logits to probability as input to do inner product with self.brackets
         # we apply softmax on the raw logits first
         softmax_input = self.softmax(input)
         # the softmax of a tensor of 0s would not be 0s anymore, so need to apply nan_targets here again to filter out
         if nan_targets is not None:
-            softmax_input[nan_targets] = 0.0
+            softmax_input = torch.masked_fill(softmax_input, nan_targets.unsqueeze(1), 0.0)
         # [batch_size, n_classes] * [n_classes] ([0, 1, 2...n_brakets-1]) -> [batch_size]
         regression_input = torch.inner(softmax_input.to(self.brackets.dtype), self.brackets.to(input.device))
         regression_loss = self.regression_loss(regression_input, target.float(), reduction=self.reduction)
@@ -82,8 +82,8 @@ class HybridCELoss(_WeightedLoss):
             regression_loss = factor1 * regression_loss * nan_targets.numel() / (num_real_targets + factor2)
 
             # set input and target with nans to -1000s for ce loss
-            input[nan_targets] = -1000
-            target[nan_targets] = -1000
+            input = torch.masked_fill(input, nan_targets.unsqueeze(1), -1000)
+            target = torch.masked_fill(target, nan_targets, -1000)
         # cross_entropy loss needs raw logits as input
         # ce_loss does not need scaling as it already ignores -1000 masked nan values
         ce_loss = F.cross_entropy(
--- a/graphium/trainer/predictor.py
+++ b/graphium/trainer/predictor.py
@@ -495,7 +495,7 @@ class PredictorModule(lightning.LightningModule):
         )
         metrics_logs = self.task_epoch_summary.get_metrics_logs()  # Dict[task, metric_logs]
         metrics_logs["_global"]["grad_norm"] = self.get_gradient_norm()
-        outputs.update(metrics_logs)  # Dict[task, metric_logs]. Concatenate them?
+        concatenated_metrics_logs.update(metrics_logs)
 
         if self.logger is not None:
             self.logger.log_metrics(
--- a/graphium/trainer/predictor.py
+++ b/graphium/trainer/predictor.py
@@ -467,7 +467,9 @@ class PredictorModule(lightning.LightningModule):
         for key in concatenated_metrics_logs:
             if isinstance(concatenated_metrics_logs[key], torch.Tensor):
                 if concatenated_metrics_logs[key].numel() > 1:
-                    concatenated_metrics_logs[key] = concatenated_metrics_logs[key].mean()
+                    concatenated_metrics_logs[key] = concatenated_metrics_logs[key][
+                        concatenated_metrics_logs[key] != 0
+                    ].mean()
 
         # If logging is skipped for this step, then log the important metrics anyway and return
         if self.skip_log_train_metrics:
--- a/graphium/trainer/predictor.py
+++ b/graphium/trainer/predictor.py
@@ -453,18 +453,11 @@ class PredictorModule(lightning.LightningModule):
         return super().on_train_batch_start(batch, batch_idx)
 
     def on_train_batch_end(self, outputs, batch: Any, batch_idx: int) -> None:
-        # Get the throughput of the batch
-        train_batch_time = time.time() - self.train_batch_start_time
-        num_graphs = self.get_num_graphs(batch["features"])
-        tput = num_graphs / train_batch_time
+        train_batch_time = time.time() - self.train_batch_start_time  # To be used for throughput calculation
 
         # Get the metrics that are logged at every step (loss, grad_norm, batch_time, batch_tput)
-        concatenated_metrics_logs = {}  # self.task_epoch_summary.concatenate_metrics_logs(metrics_logs)
+        concatenated_metrics_logs = {}
         concatenated_metrics_logs["train/loss"] = outputs["loss"]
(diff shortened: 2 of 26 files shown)


graphium/trainer/predictor.py
```diff
@@ -449,35 +447,47 @@ def flag_step(self, batch: Dict[str, Tensor], step_name: str, to_cpu: bool) -> D
 
     def on_train_batch_start(self, batch: Any, batch_idx: int) -> Optional[int]:
         self.train_batch_start_time = time.time()
+        self.is_log_train_metrics = (self.metrics_every_n_train_steps is None) or (
+            (batch_idx + 1) % self.metrics_every_n_train_steps != 0
+        )
```
zhiyil-graphcore, line 452: Does this mean if `(batch_idx + 1) % self.metrics_every_n_train_steps != 0`, `self.is_log_train_metrics` is true and we only log the metrics in `concatenated_metrics_logs` which is global loss, `batch_tput` stuff. When `(batch_idx + 1) % self.metrics_every_n_train_steps == 0`, we log `preds`, `targets` as well as individual losses? This sounds good to me.

graphium/trainer/predictor.py
```diff
@@ -449,35 +447,47 @@ def flag_step(self, batch: Dict[str, Tensor], step_name: str, to_cpu: bool) -> D
...
+            task_losses=outputs["task_losses"],
+            n_epochs=self.current_epoch,
+        )
+        metrics_logs = self.task_epoch_summary.get_metrics_logs()  # Dict[task, metric_logs]
+        metrics_logs["_global"]["grad_norm"] = self.get_gradient_norm()
+        outputs.update(metrics_logs)  # Dict[task, metric_logs]. Concatenate them?
+
         # report the training loss for each individual tasks
         for task in self.tasks:
             concatenated_metrics_logs[f"train/loss/{task}"] = outputs["task_losses"][task]
```
zhiyil-graphcore, line 493: I tested the debug config with `metrics_every_n_train_steps: 30` but somehow the individual task losses were not reported (see [here](https://wandb.ai/multitask-gnn/neurips2023_large_data_gcn_debug/runs/hss92bbw?workspace=user-zhiyi-li1230)), just wondering if there's anything missing in my setting?

graphium/trainer/predictor.py
```diff
@@ -449,35 +447,47 @@ def flag_step(self, batch: Dict[str, Tensor], step_name: str, to_cpu: bool) -> D
...
+                    concatenated_metrics_logs, step=self.global_step
+                )  # This is a pytorch lightning function call
+            return
+
+        # this code is likely repeated for validation and testing, this should be moved to a function
+        self.task_epoch_summary.update_predictor_state(
+            step_name="train",
+            targets=outputs["targets"],
+            predictions=outputs["preds"],
+            loss=outputs["loss"],  # This is the weighted loss for now, but change to task-specific loss
```
zhiyil-graphcore, line 483: Also got this error when testing `config_large_gcn.yaml`:
```
`  File "/nethome/zhiyil/graphium/graphium/ipu/ipu_wrapper.py", line 105, in on_train_batch_end
    super().on_train_batch_end(outputs, batch, batch_idx)
  File "/nethome/zhiyil/graphium/graphium/trainer/predictor.py", line 481, in on_train_batch_end
    targets=outputs["targets"],
KeyError: 'targets'
[10:29:23.668] [poptorch::python] [critical] KeyError: 'targets'`
```

graphium/trainer/predictor.py
DomInvivo: Yes exactly!

graphium/trainer/predictor.py
DomInvivo: Yeah I forgot to mention that I didn't test the code at all. It was just easier to do a wick draft than to explain in the other PR what I had in mind. 

If it's not too much to ask, can you double check everything before merging?

graphium/trainer/predictor.py
DomInvivo: See my comment below, this is a draft I have made and could have mistakes. Feel free to edit this branch before merging

graphium/trainer/predictor.py
DomInvivo: I changed `batch_idx + 1` --> `batch_idx` because I think it's a good idea to log the first batch.

graphium/trainer/predictor.py
DomInvivo: I think I fixed that in the latest commit, I moved that part before the `if self.skip_log_train_metrics`

zhiyil-graphcore: I also moved the `concatenated_metrics_logs[key] = concatenated_metrics_logs[key].mean()` part up so that we get averaged loss for the individual losses.

graphium/trainer/predictor.py
zhiyil-graphcore: I tested with debug config, seems to be fine.

zhiyil-graphcore: The test on `config_large_gcn.yaml` shows that it takes 20 minutes per epoch, which is 10 minutes longer than before...

DomInvivo: Perhaps it's the `gradient_norm` that's taking a long time?

DomInvivo: Or perhaps the logic is wrong and things gets logged more often than once every 1000 steps?

zhiyil-graphcore: I'll look into that, currently I'm not logging any metrics only the losses. Are "preds" and "targets" only used for metric logging here?

BTW, I found a way to filter 0 losses from different replica or gradient accumulation. `oncatenated_metrics_logs[key] = concatenated_metrics_logs[key].mean()` is actually averaging on r`eplica * gradien_accu * device_iter`. In `ipu_wrapper.py`, there is another averaging `outputs["loss"] = outputs["loss"].mean()`. We might just need to filter out the 0s before doing the averaging. From another point of view, as the total loss is the sum of the 5 task losses, it would not end up with zero anyway I think? We may just need to filter on individual loss for the reporting purpose.

zhiyil-graphcore: > Or perhaps the logic is wrong and things gets logged more often than once every 1000 steps?

maybe it's not that bad after all, it's taking around 22 hours for 100 epochs, compared to 19 hours previously, it is quite ok I think.

zhiyil-graphcore (approved):

graphium/trainer/predictor.py
```diff
@@ -449,43 +447,56 @@ def flag_step(self, batch: Dict[str, Tensor], step_name: str, to_cpu: bool) -> D
...
+            task_losses=outputs["task_losses"],
+            n_epochs=self.current_epoch,
+        )
+        metrics_logs = self.task_epoch_summary.get_metrics_logs()  # Dict[task, metric_logs]
+        metrics_logs["_global"]["grad_norm"] = self.get_gradient_norm()
+        outputs.update(metrics_logs)  # Dict[task, metric_logs]. Concatenate them?
+
```
luis-mueller, line 499: It seems that `outputs` is never used anywhere but also that `metrics_logs` never makes it to `concatenated_metrics_logs`, hence the extended metrics don't appear on `wandb`. @zhiyil-graphcore can you confirm this on your side?

My fix would be to simply remove the `outputs.update` call and replace with `concatenate_metrics_logs.update(metrics_logs)`. This works on my side at least to log the extended metrics every `n` steps but the main metrics at every step.

luis-mueller: If you agree I will push my change and approve.

zhiyil-graphcore: I turned off most metrics during training so didn't catch that point. Feel free to push the changes and I can verify the throughput on IPU.

DomInvivo: I have made some changes, so there might be some conflicts, please beware

graphium/trainer/predictor.py
DomInvivo: I removed the throughput computation and the `get_gradient_norm` if `self.skip_log_train_metrics is True`, so things should be a bit faster now

graphium/trainer/losses.py
```diff
@@ -63,14 +63,14 @@ def forward(self, input: Tensor, target: Tensor, nan_targets: Tensor = None) ->
 
         # set input and target with nans to 0s for regression loss
         if nan_targets is not None:
-            input[nan_targets] = 0.0
-            target[nan_targets] = 0.0
+            input = torch.masked_fill(input, nan_targets.unsqueeze(1), 0)
```
luis-mueller, line 66: These changes were necessary to enable gradient computation on my side (the previous, in-place modification caused this tedious error: `one of the variables needed for gradient computation has been modified by an in-place operation`). Not sure why the tests aren't catching this or whether this is just an issue on my side. @DomInvivo your call whether we apply this change, otherwise I can also remove it. Will definitely not be as fast and memory-friendly as the in-place but I also cannot estimate how impactful it really is to change to out-of-place (essentially creates a copy but since I am assigning it to the same variable, the garbage collector should remove the old tensor)

luis-mueller (approved):
Worked on GPU side, see comments on in-place operations in loss and my fix to actually log the metrics. LGTM

PR #411, opened 2023-07-20
filter splits NaNs

joao-alex-cunha:
There was a change introduced in commit c43cb26, to drop NaNs from spits files. However this assumes that splits files are represented as Pandas DataFrames. This is not always the case, and the code breaks when the data is represented in lists.

This PR introduces a way to filter NaNs from splits file that is compatible with both lists and pandas DataFrames.


cwognum:
Could you also add a test case related to these changes? Maybe verify each of the supported split formats actually works.

DomInvivo (approved):

PR #412, opened 2023-07-20
Draft on the changes to make the loss ignore empty batches

DomInvivo:
## Changelogs

- Removed the filtering of 0-losses, as they only change the logged loss, not the loss used in the `bacward`
- Added a function for checking how many batches are empty within an accum-batch (the combination of all accumulated batches)
- Added the parameter `loss_empty_factor` in `Predictor.compute_loss`

---

_Checklist:_

- [x] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._
- [x] _Update the API documentation if a new function is added, or an existing one is deleted._

---

- The current implementation of averaging the loss for non-empty batches, implemented by @zhiyil-graphcore in PR #391  does not work. It only changes the logged metric since the loss was already passed via the `backward` call that happens after `training_step`.
- The current implementation is an incomplete draft showing what needs to change in the dataloader to make it work.
- The current implementation only works for IPU's, since it requires pre-loading all batches of a global-batch. Current CPU/GPU implementations load batches one at a time, which makes it much more tricky to know in advance how many batches are missing.
- Not sure how much a priority this is since the problem can already be mitigated via the following measures:
  - For the datasets that are much larger, use a `epoch_fraction_sampling` that is very small, so that no dataset is entirely sparse
  - Consider using `accumulate_grad_batches > 30`, such that statistically, the number of missing batches is similar across global-batches, which means that only little noise is added

### DIFF
--- a/graphium/trainer/predictor.py
+++ b/graphium/trainer/predictor.py
@@ -233,6 +233,7 @@ class PredictorModule(lightning.LightningModule):
         loss_fun: Dict[str, Callable],
         target_nan_mask: Optional[Union[str, int]] = None,
         multitask_handling: Optional[str] = None,
+        loss_empty_factor: Optional[Dict[str, float]] = None,
     ) -> Tuple[Tensor, Dict[str, Tensor]]:
         r"""
         Compute the loss using the specified loss function, and dealing with
@@ -272,6 +273,12 @@ class PredictorModule(lightning.LightningModule):
             loss_fun:
                 Loss function to use
 
+            loss_empty_factor:
+                A dictionary used to weight the loss when there are empty batches. The loss empty factor is
+                the ratio of non-empty batches to empty batches. If all batches are empty,
+                the loss empty factor is set to 0.
+                If `None`, the loss empty factor is set to 1 for all tasks.
+
         Returns:
             Tensor:
                 weighted_loss: Resulting weighted loss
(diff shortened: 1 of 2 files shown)


zhiyil-graphcore:
One thing I'd like to get some clarification:
We compute loss with `total_loss = torch.sum(torch.stack(list(all_task_losses.values())), dim=0)`
and use the sum of 5 task losses for gradient descent, is that right?
That means the probability of getting zero loss is really low?

graphium/trainer/predictor.py
```diff
@@ -290,10 +297,16 @@ def compute_loss(
...
-        }
+
+        # Get the loss for each task
+        all_task_losses = {}
+        for task, wrapped in wrapped_loss_fun_dict.items():
+            this_loss = wrapped(preds=preds[task], target=targets[task])
+            if loss_empty_factor is not None:
+                this_loss = this_loss * loss_empty_factor[task]
+            all_task_losses[task] = this_loss
+
```
zhiyil-graphcore, line 308: We have access to the `all_task_losses` here with size of [gradient_acc*replica*device_iter], why not filter out zeros here directly?

graphium/trainer/predictor.py
```diff
@@ -290,10 +297,16 @@ def compute_loss(
...
+        # Sum the losses
         total_loss = torch.sum(torch.stack(list(all_task_losses.values())), dim=0)
```
zhiyil-graphcore, line 310: I think here it is stacking lists of losses:
task1: [0, 1, 2]
task2: [1.5, 2, 3.3]
task3: [0.3, 4,3, 0]
stacked: [1.8, 7, 5.3] then apply torch,sum --> 1.8 + 7 + 5.3 = 14.1
let me know if this makes sense.

DomInvivo:
> One thing I'd like to get some clarification: We compute loss with `total_loss = torch.sum(torch.stack(list(all_task_losses.values())), dim=0)` and use the sum of 5 task losses for gradient descent, is that right? That means the probability of getting zero loss is really low?

You are right, not only it is unlikely, but it will simply never happen because each dataset has at least one data point per molecule. So the total loss is never 0, unless the model is perfect. So the filtering of 0 loss makes no sense

graphium/trainer/predictor.py
DomInvivo: I think your right. We can do the filtering here. 

Also,  when the loss is 0 for a task, it gives 0 gradient only for the given task head, other task heads still have gradient.

Even if that task affects les the global gradient of the model, it is the only one affecting its own task head, which should get updated accordingly. 

So perhaps we don't need to adjust the loss the way that I've written it, and we can just keep normal behaviour for the gradient, but for the logging, we filter out 0- losses

DomInvivo:
Closing, related to issue #413

PR #414, opened 2023-07-21
Finetuning pipeline

WenkelF:
@DomInvivo as discussed a first draft for the Finetuning pipeline.

Two possible pipelines:
`expts/main_run_finetuning_v1.py` (probably to be removed):
- Start with pretrained model
- Remove redundant task heads
- Drop/add layers of model part to be finetuned
- Drawback: Difficult updating of hyperparameters
`expts/main_run_finetuning_v2.py`:
- Initialize new model with final finetuning architecture
- Overwrite parameters that are shared with pretrained model

Main TODOs:
- Use hydra to easy finetuning config
- Create unit test with dummy pretrained model

### DIFF
--- a/expts/finetune_configs/config_toy_finetuning_v2.yaml
+++ b/expts/finetune_configs/config_toy_finetuning_v2.yaml
@@ -1,0 +1,259 @@
+# Testing the gcn model with the PCQMv2 dataset on IPU.
+constants:
+  name: &name finetune_lipophilicity_astrazeneca
+  seed: &seed 42
+  raise_train_error: true   # Whether the code should raise an error if it crashes during training
+
+accelerator:
+  type: gpu  # cpu or ipu or gpu
+  float32_matmul_precision: medium
+  config_override:
+    datamodule:
+      args:
+        # Data handling-related
+        batch_size_training: 200
+        batch_size_inference: 200
+    trainer:
+      trainer:
+        precision: 32
+        accumulate_grad_batches: 1
+
+datamodule:
+  module_type: "ADMETBenchmarkDataModule"
+  args:
+    # TDC specific
+    tdc_benchmark_names: [lipophilicity_astrazeneca]
+    tdc_train_val_seed: *seed
+    # Featurization
+    prepare_dict_or_graph: pyg:graph
+    featurization_n_jobs: 0
+    processed_graph_data_path: "/home/mila/f/frederik.wenkel/scratch/data/graphium/datacache/admet/"
+    featurization_progress: True
+    featurization_backend: "loky"
+    processed_graph_data_path: "../datacache/tdc-admet-demo/"
+    featurization:
+      atom_property_list_onehot: [atomic-number, group, period, total-valence]
+      atom_property_list_float: [degree, formal-charge, radical-electron, aromatic, in-ring]
+      edge_property_list: [bond-type-onehot, stereo, in-ring]
+      add_self_loop: False
+      explicit_H: False # if H is included
+      use_bonds_weights: False
+      pos_encoding_as_features:
+        pos_types:
+          lap_eigvec:
+            pos_level: node
+            pos_type: laplacian_eigvec
+            num_pos: 8
+            normalization: "none" # nomrlization already applied on the eigen vectors
+            disconnected_comp: True # if eigen values/vector for disconnected graph are included
+          lap_eigval:
+            pos_level: node
+            pos_type: laplacian_eigval
+            num_pos: 8
+            normalization: "none" # nomrlization already applied on the eigen vectors
+            disconnected_comp: True # if eigen values/vector for disconnected graph are included
+          rw_pos: # use same name as pe_encoder
+            pos_level: node
+            pos_type: rw_return_probs
+            ksteps: 16
+
+    # cache_data_path: .
+    num_workers: 4 # -1 to use all
+    persistent_workers: False # if use persistent worker at the start of each epoch.
+    # Using persistent_workers false might make the start of each epoch very long.
+    featurization_backend: "loky"
+
+
+architecture:
+  model_type: FullGraphMultiTaskNetwork
+  mup_base_path: null
+  pre_nn:   # Set as null to avoid a pre-nn network
+    out_dim: 64
+    hidden_dims: 256
+    depth: 2
+    activation: relu
+    last_activation: none
+    dropout: &dropout 0.18
+    normalization: &normalization layer_norm
+    last_normalization: *normalization
+    residual_type: none
+
+  pre_nn_edges: null   # Set as null to avoid a pre-nn network
+
+  pe_encoders:
+    out_dim: 32
+    pool: "sum" #"mean" "max"
+    last_norm: None #"batch_norm", "layer_norm"
+    encoders: #la_pos |  rw_pos
+      la_pos:  # Set as null to avoid a pre-nn network
+        encoder_type: "laplacian_pe"
+        input_keys: ["laplacian_eigvec", "laplacian_eigval"]
+        output_keys: ["feat"]
+        hidden_dim: 64
+        out_dim: 32
+        model_type: 'DeepSet' #'Transformer' or 'DeepSet'
+        num_layers: 2
+        num_layers_post: 1 # Num. layers to apply after pooling
+        dropout: 0.1
+        first_normalization: "none" #"batch_norm" or "layer_norm"
+      rw_pos:
+        encoder_type: "mlp"
+        input_keys: ["rw_return_probs"]
+        output_keys: ["feat"]
+        hidden_dim: 64
+        out_dim: 32
+        num_layers: 2
+        dropout: 0.1
+        normalization: "layer_norm" #"batch_norm" or "layer_norm"
+        first_normalization: "layer_norm" #"batch_norm" or "layer_norm"
+
+
+
+  gnn:  # Set as null to avoid a post-nn network
+    in_dim: 64 # or otherwise the correct value
+    out_dim: &gnn_dim 96
+    hidden_dims: *gnn_dim
+    depth: 4
+    activation: gelu
+    last_activation: none
+    dropout: 0.1
+    normalization: "layer_norm"
+    last_normalization: *normalization
+    residual_type: simple
+    virtual_node: 'none'
+    layer_type: 'pyg:gcn' #pyg:gine #'pyg:gps' # pyg:gated-gcn, pyg:gine,pyg:gps
+    layer_kwargs: null # Parameters for the model itself. You could define dropout_attn: 0.1
+
+  graph_output_nn:
+    graph:
+      pooling: [sum]
+      out_dim: *gnn_dim
+      hidden_dims: *gnn_dim
+      depth: 1
+      activation: relu
+      last_activation: none
+      dropout: *dropout
+      normalization: *normalization
+      last_normalization: "none"
+      residual_type: none
+
+  task_heads:
+    qm9:
+      task_level: graph
+      out_dim: 19
+      hidden_dims: 128
+      depth: 2
+      activation: relu
+      last_activation: none
+      dropout: *dropout
+      normalization: *normalization
+      last_normalization: "none"
+      residual_type: none
+    tox21:
+      task_level: graph
+      out_dim: 12
+      hidden_dims: 64
+      depth: 2
+      activation: relu
+      last_activation: sigmoid
+      dropout: *dropout
+      normalization: *normalization
+      last_normalization: "none"
+      residual_type: none
+    zinc:
+      task_level: graph
+      out_dim: 3
+      hidden_dims: 32
+      depth: 2
+      activation: relu
+      last_activation: none
+      dropout: *dropout
+      normalization: *normalization
+      last_normalization: "none"
+      residual_type: none
+
+#Task-specific
+predictor:
+  metrics_on_progress_bar:
+    lipophilicity_astrazeneca: ["mae"]
+  loss_fun:
+    lipophilicity_astrazeneca: mae
+  random_seed: *seed
+  optim_kwargs:
+    lr: 4.e-5 # warmup can be scheduled using torch_scheduler_kwargs
+    # weight_decay: 1.e-7
+  torch_scheduler_kwargs:
+    module_type: WarmUpLinearLR
+    max_num_epochs: &max_epochs 10
+    warmup_epochs: 2
+    verbose: False
+  scheduler_kwargs:
+  #  monitor: &monitor qm9/mae/train
+  #  mode: min
+  #  frequency: 1
+  target_nan_mask: null # null: no mask, 0: 0 mask, ignore-flatten, ignore-mean-per-label
+  multitask_handling: flatten # flatten, mean-per-label
+
+# Task-specific
+metrics:
+  lipophilicity_astrazeneca:
+    - name: mae
+      metric: mae
+      target_nan_mask: null
+      multitask_handling: flatten
+      threshold_kwargs: null
+    - name: spearman
+      metric: spearmanr
+      threshold_kwargs: null
+      target_nan_mask: null
+      multitask_handling: mean-per-label
+    - name: pearson
+      metric: pearsonr
+      threshold_kwargs: null
+      target_nan_mask: null
+      multitask_handling: mean-per-label
+    - name: r2_score
+      metric: r2
+      target_nan_mask: null
+      multitask_handling: mean-per-label
+      threshold_kwargs: null
+
+trainer:
+  seed: *seed
+  logger:
+    save_dir: logs/finetuned_models/
+    name: *name
+    project: *name
+  #early_stopping:
+  #  monitor: *monitor
+  #  min_delta: 0
+  #  patience: 10
+  #  mode: &mode min
+  finetuning:
+    task: lipophilicity_astrazeneca
+    level: graph
+    module: task_heads              # options are (pre_nn, pre_nn_edges) gnn, graph_output_nn, task_heads
+    choice: zinc                    # choose one from pretrained model (here qm9, tox12 or zinc)
+    drop_depth: 1                   # how many layers to drop at the end:
+                                    # must be smaller than pretrained model's depth
+                                    # for TaskHeads:
+                                    #   ==0 only if finetune task has same out_dim
+                                    #   >=1 otherwise, as we need to update out_dim  
+    add_finetune_depth: 0           # if desired, we can also finetune layers before added layers
+    added_layers:                   # here we simply extend the layer, we finetune from
+      out_dim: 1
+      depth: 3
+    
+    epoch_unfreeze_all: none
+
+  model_checkpoint:
+    dirpath: finetuned_models/
+    filename: finetuned-toy-model-{epoch}
+    # monitor: *monitor
+    # mode: *mode
+    # save_top_k: 1
+    save_last: True
+  trainer:
+    max_epochs: *max_epochs
+    min_epochs: 1
+    check_val_every_n_epoch: 2
--- a/expts/main_run_finetuning_v2.py
+++ b/expts/main_run_finetuning_v2.py
@@ -1,0 +1,153 @@
+# General imports
+import argparse
+import os
+from os.path import dirname, abspath
+import yaml
+from copy import deepcopy
+from omegaconf import DictConfig
+import timeit
+from loguru import logger
+from datetime import datetime
+import torch.nn as nn
+from lightning.pytorch.utilities.model_summary import ModelSummary
+
+from graphium.utils.mup import set_base_shapes
+from graphium.finetuning import modify_cfg_for_finetuning, MolecularFinetuning
+
+# Current project imports
+import graphium
+from graphium.config._loader import (
+    load_datamodule,
+    load_metrics,
+    load_architecture,
+    load_predictor,
+    load_trainer,
+    save_params_to_wandb,
+    load_accelerator,
+    load_yaml_config,
+)
+from graphium.utils.safe_run import SafeRun
+from graphium.utils.command_line_utils import update_config, get_anchors_and_aliases
+
+
+# WandB
+import wandb
+
+# Set up the working directory
+MAIN_DIR = dirname(dirname(abspath(graphium.__file__)))
+
+CONFIG_FILE = "expts/finetune_configs/config_toy_finetuning_v2.yaml"
+
+os.chdir(MAIN_DIR)
+
+
+def main(cfg: dict, run_name: str = "main", add_date_time: bool = True) -> None:
+    st = timeit.default_timer()
+
+    date_time_suffix = ""
+    if add_date_time:
+        date_time_suffix = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
+
+
+
+    ########################
+    # Changes for finetuning
+
+    # Modify config
+    cfg = modify_cfg_for_finetuning(cfg)
+
+    ########################
+
+
+    wandb.init(entity="multitask-gnn", project=cfg["constants"]["name"], config=cfg)
+
+    # Initialize the accelerator
+    cfg, accelerator_type = load_accelerator(cfg)
+
+    # Load and initialize the dataset
+    datamodule = load_datamodule(cfg, accelerator_type)
+
+    # Initialize the network
+    model_class, model_kwargs = load_architecture(
+        cfg,
+        in_dims=datamodule.in_dims,
+    )
+
+    datamodule.prepare_data()
+
+    metrics = load_metrics(cfg)
+    logger.info(metrics)
+
+    predictor = load_predictor(
+        cfg, model_class, model_kwargs, metrics, accelerator_type, datamodule.task_norms
+    )
+
+
+    ########################
+    # Changes for finetuning
+
+    # Load pretrained & replace in predictor
+    pretrained_model = predictor.load_pretrained_models('dummy-pretrained-model').model     # make pretrained model part of config
+    
+    # Adapt pretrained model to new task
+    # We need to overwrite shared weights with pretrained
+
+    predictor.model.overwrite_with_pretrained(cfg, pretrained_model)
+
+    # (Un)freezing will be handled by finetuning callback added to trainer
+
+    predictor.model = set_base_shapes(predictor.model, base=None)       # how do we deal with muP for finetuning; it
+    ########################
+
+
+    logger.info(predictor.model)
+    logger.info(ModelSummary(predictor, max_depth=4))
+
+
+    trainer = load_trainer(cfg, run_name, accelerator_type, date_time_suffix)
+    
+    ########################
+    # Changes for finetuning
+
+    # Add the pl.BaseFinetuning callback to trainer
+    
+    trainer.callbacks.append(
+        MolecularFinetuning(cfg)
+    )
+    ########################
+
+    save_params_to_wandb(trainer.logger, cfg, predictor, datamodule)
+
+    # Determine the max num nodes and edges in training and validation
+    logger.info("About to set the max nodes etc.")
+    predictor.set_max_nodes_edges_per_graph(datamodule, stages=["train", "val"])
+
+    # Run the model training
+    with SafeRun(name="TRAINING", raise_error=cfg["constants"]["raise_train_error"], verbose=True):
+        trainer.fit(model=predictor, datamodule=datamodule)
+
+    # Determine the max num nodes and edges in testing
+    predictor.set_max_nodes_edges_per_graph(datamodule, stages=["test"])
+
+    # Run the model testing
+    with SafeRun(name="TESTING", raise_error=cfg["constants"]["raise_train_error"], verbose=True):
+        trainer.test(model=predictor, datamodule=datamodule)  # , ckpt_path=ckpt_path)
+
+    logger.info("--------------------------------------------")
+    logger.info("total computation used", timeit.default_timer() - st)
+    logger.info("--------------------------------------------")
+    wandb.finish()
+
+    return trainer.callback_metrics
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config", help="Path to the config file", default=None)
+
+    args, unknown_args = parser.parse_known_args()
+    if args.config is not None:
+        CONFIG_FILE = args.config
+    cfg = load_yaml_config(CONFIG_FILE, MAIN_DIR, unknown_args)
+
+    main(cfg)
--- a/graphium/data/datamodule.py
+++ b/graphium/data/datamodule.py
@@ -1876,19 +1876,23 @@ class MultitaskFromSmilesDataModule(BaseDataModule, IPUDataModuleModifier):
             # Split from an indices file
             file_type = self._get_data_file_type(splits_path)
 
+            train, val, test = split_names
+
             if file_type == "pt":
                 splits = torch.load(splits_path)
+                train_indices = np.asarray(splits[train]).astype("int").tolist()
+                val_indices = np.asarray(splits[val]).astype("int").tolist()
+                test_indices = np.asarray(splits[test]).astype("int").tolist()
             elif file_type in ["csv", "tsv"]:
                 with fsspec.open(str(splits_path)) as f:
                     splits = self._read_csv(splits_path)
+                train_indices = np.asarray(splits[train].dropna()).astype("int").tolist()
+                val_indices = np.asarray(splits[val].dropna()).astype("int").tolist()
+                test_indices = np.asarray(splits[test].dropna()).astype("int").tolist()
             else:
                 raise ValueError(
                     f"file type `{file_type}` for `{splits_path}` not recognised, please use .pt, .csv or .tsv"
                 )
-            train, val, test = split_names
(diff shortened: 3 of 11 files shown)


expts/main_run_finetuning_v2.py
```diff
@@ -0,0 +1,153 @@
...
+    if add_date_time:
+        date_time_suffix = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
+
+
+
+    ########################
+    # Changes for finetuning
+
+    # Modify config
+    cfg = modify_cfg_for_finetuning(cfg)
```
DomInvivo, line 57: Look into hydra

expts/main_run_finetuning_v2.py
```diff
@@ -0,0 +1,153 @@
...
+        in_dims=datamodule.in_dims,
+    )
+
+    datamodule.prepare_data()
+
+    metrics = load_metrics(cfg)
+    logger.info(metrics)
+
+    predictor = load_predictor(
+        cfg, model_class, model_kwargs, metrics, accelerator_type, datamodule.task_norms
```
DomInvivo, line 82: I would copy the `model_kwargs` from the `pretrained_model`, but modify the configs with a similar logic to `modify_cfg_for_pretrained`

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,56 @@
...
+        Finetuning logic
+        """
+        super().__init__()
+
+        cfg_finetune = cfg['trainer']['finetuning']
+        self.depth = cfg_finetune['added_layers']['depth']
+        try:
+            self.depth += cfg_finetune['add_finetune_depth']
+        except:
+            pass
```
DomInvivo, line 26: I guess this is in case `'add_finetune_depth'` is missing? You an do this instead, which puts a default value of 0 if the key is missing. But I would suggest changing the `__init__` altogether
```suggestion
        self.depth += cfg_finetune.get("add_finetune_depth", 0)
```

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,56 @@
...
+import pytorch_lightning as pl
+
+from torch.optim.optimizer import Optimizer
+from pytorch_lightning.callbacks import BaseFinetuning
+
+
+class MolecularFinetuning(BaseFinetuning):
+    def __init__(
+        self,
+        cfg,
```
DomInvivo, line 12:
```suggestion
        added_depth: int,
        on_task: str,
        epoch_unfreeze_all: int,
        add_finetune_depth: int = 0,
        ...
```

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,56 @@
+from typing import Iterable, List, Dict, Tuple, Union, Callable, Any, Optional, Type
+
+import pytorch_lightning as pl
+
+from torch.optim.optimizer import Optimizer
+from pytorch_lightning.callbacks import BaseFinetuning
+
+
+class MolecularFinetuning(BaseFinetuning):
```
DomInvivo, line 9: Nothing from the class suggests that it's for Molecules only. I would cal it something more generic, like `GraphFinetuning` or `FullGraphNetworkFinetuning` (if specific to the `FullGraphNetwork` class, which it seems to be since it's using `model.task_heads` somewhere)

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,56 @@
...
+        train_bn: bool = False
+
+    ):
+        """
+        Finetuning logic
+        """
+        super().__init__()
+
+        cfg_finetune = cfg['trainer']['finetuning']
+        self.depth = cfg_finetune['added_layers']['depth']
```
DomInvivo, line 22: Add some checks in the `__init__` that `depth` and `add_fine_depth` and `on_task` are consistent with the model. Perhaps this would require to pass the model at the `__init__`, which is weird, but we want the code to crash BEFORE we start processing the data and training the model. We want a quick crash, instead of waiting 20min+ before it crashes for a wrong depth.

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,56 @@
...
+
+from torch.optim.optimizer import Optimizer
+from pytorch_lightning.callbacks import BaseFinetuning
+
+
+class MolecularFinetuning(BaseFinetuning):
+    def __init__(
+        self,
+        cfg,
+        train_bn: bool = False
```
DomInvivo, line 13: what is `train_bn`? Name seems ambiguous

graphium/finetuning/utils.py
```diff
@@ -0,0 +1,27 @@
+from copy import deepcopy
+
+
+def modify_cfg_for_finetuning(cfg):
```
DomInvivo, line 4: This could be a function within `FeedForwardNN` and `FullGraphNetwork`

graphium/nn/architectures/global_architectures.py
```diff
@@ -1278,6 +1301,95 @@ def set_max_num_nodes_edges_per_graph(self, max_nodes: Optional[int], max_edges:
 
         self.task_heads.set_max_num_nodes_edges_per_graph(max_nodes, max_edges)
 
+    def modify_architecture(self, cfg):
```
DomInvivo, line 1304: Pass the parameters here, not the configs.
```suggestion
    def modify_architecture(self, choice, task, add_layers, out_dim, ...):
```

Also, as discussed, `choice` is an ambiguous name

graphium/nn/architectures/global_architectures.py
```diff
@@ -1278,6 +1301,95 @@ def set_max_num_nodes_edges_per_graph(self, max_nodes: Optional[int], max_edges:
...
+            in_dim = self.task_heads.task_heads[task].layers[-1].out_dim
+            added_layers = FeedForwardNN(in_dim, **cfg_finetune['added_layers']).layers
+
+            self.task_heads.task_heads[task].add_layers(added_layers)
+
+            self.task_heads.task_heads_kwargs = {task: self.task_heads.task_heads_kwargs[choice]}
+            self.task_heads.task_heads_kwargs[task].update(
+                {
+                    'out_dim': cfg_finetune['added_layers']['out_dim'],
+                    'depth': self.task_heads.task_heads_kwargs[task]['depth'] + cfg_finetune['added_layers']['depth'],
```
DomInvivo, line 1324: Be careful if some values are optional.

DomInvivo (commented):
Here's my review. It mostly aligns with what we discussed last time

expts/finetune_configs/config_toy_finetuning_v2.yaml
```diff
@@ -0,0 +1,259 @@
...
+    add_finetune_depth: 0           # if desired, we can also finetune layers before added layers
+    added_layers:                   # here we simply extend the layer, we finetune from
+      out_dim: 1
+      depth: 3
+    
+    epoch_unfreeze_all: none
+
+  model_checkpoint:
+    dirpath: finetuned_models/
+    filename: finetuned-toy-model-{epoch}
```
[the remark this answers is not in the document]
cwognum: It's a placeholder that will be used to format the string. See [`ModelCheckpoint`'s `filename` parameter](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html). Similar to normal string formatting:

```python
s = "{epoch}_text"
s.format(epoch=10)
# Returns: '10_text'

```

graphium/nn/architectures/global_architectures.py
```diff
@@ -1278,6 +1301,95 @@ def set_max_num_nodes_edges_per_graph(self, max_nodes: Optional[int], max_edges:
...
+
+    def overwrite_complete_module(self, module, pretrained_model):
+        """Completely overwrite the specified module
+        """
+        if module == "pre_nn":
+            try:
+                self.pre_nn.layers = pretrained_model.pre_nn.layers
+            except:
+                logger.warning(f"Pretrained ({pretrained_model.pre_nn}) and/or finetune model ({self.pre_nn}) do not use a pre-NN.")
+                pass
```
cwognum, line 1377: Just a warning? This seems like a major configuration error and should throw an Exception? 

(FYI:  `pass` is not needed here)

graphium/nn/architectures/global_architectures.py
```diff
@@ -1278,6 +1301,95 @@ def set_max_num_nodes_edges_per_graph(self, max_nodes: Optional[int], max_edges:
...
+
+        cfg_finetune = cfg['finetuning']
+        task_head_from_pretrained = cfg_finetune['task_head_from_pretrained']
+        task = cfg_finetune['task']
+        added_depth = cfg_finetune['added_depth']
+
+        for module in ['pre_nn', 'pre_nn_edges', 'gnn', 'graph_output_nn', 'task_heads']:
+
+            if module == cfg_finetune['module_from_pretrained']:
+                break
```
cwognum, line 1342: `break` (exit the loop entirely) or `continue` (skip a single loop iteration)?

cwognum (commented):
Hi @WenkelF, sorry for the delay here!

Thank you for this first implementation! 

I left comments whenever something stood out to me, but I'm aware that this PR is still WIP. Sorry if I pointed out some things that you were already planning to change. 

It would be super helpful if you could document the main fine-tuning "flow". Furthermore, I would suggest to simplify the process by adding support for one feature at a time, instead of having half-implemented features. This will make understanding, debugging, testing and maintaining the code base a lot easier. 

Happy to help next week!

graphium/finetuning/finetuning.py
DomInvivo: Ahhh I see! `train_bn` as in "should we continue training the bach_norm". Makes sense.

expts/hydra-configs/dataset/lipophilicity.yaml
```diff
@@ -0,0 +1,309 @@
...
+      normalization: *normalization
+      last_normalization: "none"
+      residual_type: none
+    zinc:
+      task_level: graph
+      out_dim: 3
+      hidden_dims: 32
+      depth: 2
+      activation: relu
+      last_activation: none
```
DomInvivo, line 103: This doesn't make sense. We should not have any architectural choice from the original pre-trained model in here. Only things that would change.

That way, we can take different pre-trained models that have different hparams/seed and fine-tune them all with the same file.

expts/main_run_finetuning.py
```diff
@@ -0,0 +1,131 @@
...
+from graphium.utils.safe_run import SafeRun
+
+import hydra
+
+# WandB
+import wandb
+
+# Set up the working directory
+MAIN_DIR = dirname(dirname(abspath(graphium.__file__)))
+os.chdir(MAIN_DIR)
```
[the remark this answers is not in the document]
DomInvivo: The issues you're mentioning for this file are very related to the issues from `main_run_multitask.py`, which is what this file is based on. Perhaps you could open a new PR where you change the path handling and make sure it works?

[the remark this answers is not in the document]
DomInvivo: And perhaps take care of the wandb config at the same time?

graphium/data/datamodule.py
```diff
@@ -1959,8 +1961,11 @@ def get_data_hash(self):
         # so that the data cache does not need to be regenerated
         # when epoch_sampling_fraction has changed.
         for task in self.task_specific_args.keys():
-            if "epoch_sampling_fraction" in args[task].keys():
-                args[task].pop("epoch_sampling_fraction")
+            try:
+                if "epoch_sampling_fraction" in args[task].keys():
+                    args[task].pop("epoch_sampling_fraction")
+            except:
+                pass
```
DomInvivo, line 1968: I don't even understand the point of having a `try` and an `if` there. `Dict.pop` works even if the key is not available.

But we need to make sure that `args[task]` is not used elsewhere, even outside the current function since dict are passed as pointers. We only want to remove `epoch_sampling` for the hash key. So I would suggest the following.

```suggestion
            args[task] = deepcopy(args[task])
            args[task].pop("epoch_sampling_fraction")
```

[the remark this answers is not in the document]
DomInvivo: See suggestion below.

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,72 @@
+from typing import Iterable, List, Dict, Tuple, Union, Callable, Any, Optional, Type
+
+import pytorch_lightning as pl
+
+from torch.optim.optimizer import Optimizer
+from pytorch_lightning.callbacks import BaseFinetuning
+
+
+class GraphFinetuning(BaseFinetuning):
+    def __init__(self, cfg, train_bn: bool = False):
```
DomInvivo, line 10: Change to explicitly pass parameters. 
```suggestion
    def __init__(self, fine-tuning, architecture, module_from_pretrained, ....................., train_bn: bool = False):
```

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,72 @@
...
+            modules = pl_module.model.pre_nn
+        elif module == "pre_nn_edges":
+            modules = pl_module.model.pre_nn_edges
+        elif module == "gnn":
+            modules = pl_module.model.gnn
+        elif module == "graph_output_nn":
+            modules = pl_module.model.task_heads.graph_output_nn
+        elif module == "task_heads":
+            modules = pl_module.model.task_heads.task_heads
+
```
DomInvivo, line 57: `else: raise "Wrong module"`

graphium/finetuning/finetuning.py
```diff
@@ -0,0 +1,72 @@
...
+        elif module == "pre_nn":
+            modules = pl_module.model.pre_nn
+        elif module == "pre_nn_edges":
+            modules = pl_module.model.pre_nn_edges
+        elif module == "gnn":
+            modules = pl_module.model.gnn
+        elif module == "graph_output_nn":
+            modules = pl_module.model.task_heads.graph_output_nn
+        elif module == "task_heads":
+            modules = pl_module.model.task_heads.task_heads
```
DomInvivo, line 56: I would define all these in a dictionary `_module_map = {pe_encoders: pl_module.model.encoder_manager, ...}` directly in the `__init__`. That way, with inheritance, someone could modified the entries without copy-pasting all the logic.

`_module_map` can replace the `module_list` you already have.

But instead of a regular dict, using an `OrderedDict` would also allow you to say something like: "freeze everything before task_heads" in a very simple way.

graphium/finetuning/finetuning_architecture.py
DomInvivo, line 1: It's a bad idea to have a `FullGraphFinetuningNetwork` that basically copy-pastes most of the functionality of `FullGraphNetwork`.

Either use inheritance, or implement the fine-tuning logic directly within `FullGraphNetwork`

graphium/nn/architectures/global_architectures.py
```diff
@@ -272,6 +273,28 @@ def _create_layers(self):
             if ii < len(residual_out_dims):
                 this_in_dim = residual_out_dims[ii]
 
+    def drop_layers(self, depth: int) -> None:
+        r"""
+        Remove the last layers of the model part.
+        """
+
+        assert depth >= 0
+        assert depth <= len(self.layers)
```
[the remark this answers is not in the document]
DomInvivo: hmmm... We've used tons of `assert` everywhere in the code. Do you recommend doing that everytime:
```python
if depth < 0: raise "depth should be >= 0"
```

DomInvivo (commented):
Don't forget that the objective is to release a working, incomplete version first. Then refining it with more complex fine-tuning possibilities.

graphium/nn/architectures/global_architectures.py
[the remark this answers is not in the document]
cwognum: Yeah, I would. My understanding is that assert statements can be removed at compile time (to improve performance, which seems specifically relevant to Graphium), so any error that a downstream user can make and should be informed about should ideally be a raised and specific Exception. To have sanity checks during debugging / development, you can still use assert statements.

WenkelF:
@DomInvivo thanks for your comments. Here is also a quick overview of the updates

Updates:
- Moved finetuning-related architecture out of FullGraphMultitaskNetwork to FullGraphFinetuningNetwork under graphium/finetuning/finetuning_architecture
-- More flexibility to build finetuning head on top of pretrained model
-- For now, finetuning is specific to FullGraphMultitaskNetwork (as pretrained model), but we can finetune from gnn, graph_output_nn and task_heads now
-- Additionally, we can add a flexible FinetuningHead, inheriting from nn.Module and MupMixin. For example, it can be a FeedForwardNN, FeedForwardPyg, TaskHeads, or a custom network that can be added to FINETUNING_HEAD_DICT
-- FinetuningHead needs a bit more work to correctly support all possible scenarios (modules/levels to finetune from); if not used, we automatically fall back to the finetuning logic, where we can only drop & add depth to the module we finetune from
-- Finished functions for overwriting shared weights and partially freezing modules while training and extended them from only task_heads to also gnn and graph_output_nn
- Added preliminary unit test the discussed “minimal” finetuning pipeline (finetune from one of the task heads)
-- We currently test depths, in_dim, ou_dim of changed modules are correct and overwritten correctly
-- Test does not cover training/freezing yet
-- Planning to add tests for finetuning from other modules as well
- Configuration has been integrated with hydra but needs more work to be easier to use (@luis-mueller is helping with that)

Remarks:
- Hidden_dims can only be used as int together with depth at the moment
- Currently, everything has only been tested on cpu/gpu

graphium/data/datamodule.py
WenkelF: Indeed, this is not an ideal fix. I wanted to investigate the issue a bit more.

We cannot use `args[task].pop("epoch_sampling_fraction")` because `args[task]` may be of class `DatasetProcessingParams` instead of `Dict`. In particular, `ADMETBenchmarkDataModule` makes use of `DatasetProcessingParams`.

This originates from changes in 4b82ba3bb049b7ac1376f93e9537bd1a5d3842e0. We were using a Dict in all configs I saw, although I see a comment `# To be replaced by a new class "DatasetParams"` everywhere it appears.

Will create issue and think about a fix.

expts/main_run_finetuning.py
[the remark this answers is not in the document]
WenkelF: Thanks for the comments @cwognum. I created an issue related to the working directory and wandb.

expts/main_run_finetuning.py
```diff
@@ -0,0 +1,131 @@
...
+    cfg = modify_cfg_for_finetuning(cfg)
+
+    run_name: str = "main"
+    add_date_time: bool = True
+
+    st = timeit.default_timer()
+
+    date_time_suffix = ""
+    if add_date_time:
+        date_time_suffix = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
```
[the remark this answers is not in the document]
WenkelF: I also added this part to the issue mentioned above.

graphium/nn/architectures/global_architectures.py
WenkelF: Yes, because we only want to fully overwrite modules BEFORE the module we finetune from. The finetuning module is overwritten partially and modules afterwards are not overwritten (they keep their random initialization).

graphium/finetuning/finetuning.py
WenkelF: Addressed in 76e2ba6777935056030eefb9f82dfc0b53a3884d

graphium/finetuning/finetuning.py
WenkelF: Addressed in 76e2ba6777935056030eefb9f82dfc0b53a3884d

graphium/nn/architectures/global_architectures.py
```diff
@@ -1278,6 +1301,95 @@ def set_max_num_nodes_edges_per_graph(self, max_nodes: Optional[int], max_edges:
...
+            self.task_heads.task_heads_kwargs[task].update(
+                {
+                    'out_dim': cfg_finetune['added_layers']['out_dim'],
+                    'depth': self.task_heads.task_heads_kwargs[task]['depth'] + cfg_finetune['added_layers']['depth'],
+                    'name': 'NN-' + task
+                }
+            )
+
+        else:
+            raise NotImplementedError()
```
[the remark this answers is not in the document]
WenkelF: I had two different finetuning approaches (v1 & v2) implemented in the beginning. This is a relic of v1 and, which we will not be using. I removed the code by now.

graphium/nn/architectures/global_architectures.py
WenkelF: Similar as above, this was part of finetuning v1, which we are not using. I have not removed the function because a similar function was already implemented by someone else (although it was not used anywhere).

Do you recommend removing the part of the code?

expts/hydra-configs/dataset/lipophilicity.yaml
WenkelF: I agree. The configurations are still structured in a way where we have access to both the full config of the pretrained model and the pretraining-related config. And the modify_cfg_for_finetuning function consolidates information to one config.

This will be fixed once we incorporate the new hydra config from #421. We will still need modify_cfg_for_finetuning as of now. Therefore, it could be good waiting for the final version.

graphium/finetuning/utils.py
WenkelF: I am not sure, we can remove the function and replace inside the networks. As discussed in that #411, it might be possible to have a similar function within load_architecture in `graphium/data/_loader.py` but depends on the PR.

graphium/finetuning/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+    cfg_finetune = cfg['finetuning']
+    module = cfg_finetune['module_from_pretrained']
+    task = cfg_finetune['task']
+    level = cfg_finetune['level']
+    task_head_from_pretrained = cfg_finetune['task_head_from_pretrained']
```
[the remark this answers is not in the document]
cwognum: Fyi: You don't need the None here. That's the default value.

graphium/nn/architectures/global_architectures.py
cwognum: Ah check! I understand now, thanks!

graphium/data/datamodule.py
WenkelF, line 1: @cwognum, could the issue in `expts/test_multitask_datamodue.py` be caused by this recent change?

cwognum:
xref: https://github.com/WenkelF/graphium/pull/10

WenkelF:
This pull introduces the following:
- Finetuning pipeline
- Updates to hydra configuration (by @cwognum )

The finetuning pipeline is maintained separately from existing architectures under `graphium/finetuning/finetuning_architecture`. The FullGraphFinetuningNetwork comes with two submodules:
- `PretrainedModel`
- `FinetuningHead `(optional)

`PretrainedModel` can load pretrained models (e.g., FullGraphMultitaskNetwork) and allows basic finetuning operations (in particular from the task heads), including dropping or extending layers of the module to finetune from.

`FinetuningHead` is optional and allows to define fully customizable networks that are applied on top of PretrainedModel. If not specified, we fall back to basic finetuning explained above.

Training is handled by the `GraphFinetuning `callback in `graphium/finetuning/finetuning`.

All methods in [graphium/finetuning](https://github.com/WenkelF/graphium/tree/explore_finetuning/graphium/finetuning) are implemented such that they are not specific to a pretrained model or finetuning head. This is achieved by requiring the pretrained model to come with a module_map (see, e.g., `FullGraphMultitaskNetwork`) that facilitates setting pretrained weights and freezing the correct layers.

The new unit test `tests/test_finetuning` tests the pipeline end-to-end for a specific case. Together with the corresponding config, it might be a good starting point to understand the pipeling and what information is needed in the config.

The updates to hydra allow to easily switch between benchmarking and finetuning. Major changes are documented here https://github.com/WenkelF/graphium/pull/10

graphium/trainer/predictor.py
```diff
@@ -96,20 +95,46 @@ def __init__(
...
             eval_options[task].check_metrics_validity()
 
+        # Work-around to retain task level when model_kwargs are modified for FullGraphFinetuningNetwork
+        if "task_heads_kwargs" in model_kwargs.keys():
+            task_heads_kwargs = model_kwargs["task_heads_kwargs"]
+        elif "pretrained_model_kwargs" in model_kwargs.keys():
+            # This covers finetuning cases where we finetune from the task_heads
+            task_heads_kwargs = model_kwargs["pretrained_model_kwargs"]["task_heads_kwargs"]
+        else:
+            raise ValueError("incorrect model_kwargs")
```
DomInvivo, line 105: I don't think this should be here. I think that, if you are using a pre-trained model, you should pass directly `model_kwargs["pretrained_model_kwargs"]` into `model_kwargs`.

graphium/trainer/predictor.py
```diff
@@ -96,20 +95,46 @@ def __init__(
...
+        self.task_heads_kwargs = task_heads_kwargs
+
         self._eval_options_dict: Dict[str, EvalOptions] = eval_options
         self._eval_options_dict = {
             self._get_task_key(
-                task_level=model_kwargs["task_heads_kwargs"][key]["task_level"], task=key
+                task_level=task_heads_kwargs[key]["task_level"],
+                task=key
+                # task_level=model_kwargs["task_heads_kwargs"][key]["task_level"], task=key
```
DomInvivo, line 113: In general, the `PredictorModule` should be agnostic to the model passed. By having the `self._get_task_key` here, it forces a certain architecture in the config which is not very flexible. 

I see that this logic was introduced already in the code prior to this PR. If it requires too many changes, let's open a new issue.

graphium/trainer/predictor.py
```diff
@@ -96,20 +95,46 @@ def __init__(
...
+        # We now need to define the module_map in pretrained_model in FinetuningNetwork
+        # self._module_map = OrderedDict(
+        #     pe_encoders=self.model.encoder_manager,
+        #     pre_nn=self.model.pre_nn,
+        #     pre_nn_edges=self.model.pre_nn_edges,
+        #     gnn=self.model.gnn,
+        #     graph_output_nn=self.model.task_heads.graph_output_nn,
+        #     task_heads=self.model.task_heads.task_heads,
+        # )
+
```
DomInvivo, line 132: Delete this comment.

graphium/trainer/predictor.py
```diff
@@ -96,20 +95,46 @@ def __init__(
...
         loss_fun = {
             self._get_task_key(
-                task_level=model_kwargs["task_heads_kwargs"][key]["task_level"], task=key
+                task_level=task_heads_kwargs[key]["task_level"],
+                task=key
```
DomInvivo, line 136: Again, don't like how the config structure is imposed. Perhaps `task_level` should simply be passed to the `PredictorModule` to keep flexibility

tests/dummy-pretrained-model-gpu.ckpt
DomInvivo, line 1: Why duplicating the model for CPU and GPU? Models should be agnostic to the training hardware, and to the fine-tuning hardware. You can train on CPU and fine-tune on GPU or IPU

DomInvivo (changes requested):
Mostly looks good, great work on this major PR!
A few changes to make, and some comments.

DomInvivo:
@zhiyil-graphcore @s-maddrellmander We'll need your help here to fix the tests for IPU. And ideally, have a test that loads a CPU-trained model onto IPU for finetuning.

graphium/cli/finetune.py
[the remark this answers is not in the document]
cwognum: Why not keep it general? There is no reason this should be limited to just ADMET and I would argue against having a CLI for each experiment. That will lead to a lot of code repetition!

expts/hydra-configs/tasks/admet.yaml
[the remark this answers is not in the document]
cwognum: Not sure I understand. The purpose is the same as any `tasks` config; Allows you to include these tasks in a training/fine-tuning loop!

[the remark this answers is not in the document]
cwognum: If you're asking why we have a config to merge two subconfigs: That's mostly to remain backwards compatible. I had to split the config in two parts since we only need one part during finetuning, whereas we need both during training.

expts/hydra-configs/finetuning/admet.yaml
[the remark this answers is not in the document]
cwognum: I understand where you're coming from, but I personally don't think renaming is the solution. If you would create a config for each task, you end up with ~20 configs with lots of repetition. I would keep the name general, but make the variability more explicit with comments in the yaml or even by introducing new variables in the config.

env.yml
```diff
@@ -28,7 +28,7 @@ dependencies:
   - gcsfs >=2021.6
 
   # ML packages
-  - cudatoolkit # works also with CPU-only system.
+  - cuda-version =11.8 # works also with CPU-only system.
```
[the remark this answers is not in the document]
cwognum: Good catch! Copied this from @hadim his tutorial, but we likely don't need it!

expts/hydra-configs/finetuning/finetuning.yaml
```diff
@@ -0,0 +1,82 @@
...
+epoch_unfreeze_all: none                          # Indicate an epoch to unfreeze all layers
+
+
+
+
+
+
+# Changes to be implemented
+
+# predictor:
```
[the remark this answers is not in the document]
cwognum: This file no longer exists. Resolving the comment!

graphium/cli/finetune.py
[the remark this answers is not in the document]
DomInvivo: There are lots of things hardcoded for admet. But I do agree, once we have more things to fine-tune on, we can remove the hardcoding and use the same file.

expts/hydra-configs/tasks/admet.yaml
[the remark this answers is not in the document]
DomInvivo: Ok. Do you think that using more explicit names such as `task_heads` and `data_and_metrics` would make sense?

expts/hydra-configs/finetuning/admet.yaml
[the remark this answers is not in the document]
DomInvivo: Yeah, let's add some comments to show which parameters are usually overwritten in the other configs. Otherwise it's confusing

graphium/cli/finetune.py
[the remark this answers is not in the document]
cwognum: > There are lots of things hardcoded for admet. 

I don't think so? Only the filtering of the config, but that can easily be seen as a more general config design pattern.

graphium/trainer/predictor.py
WenkelF: You are right, thanks for pointing out. 

We could achieve this by getting the task-specific information (which is only the task level as far as I know) from the datamodule.

Here is how this could be done:
a3d471540bc3f69cc0af2346ab0e66141b65311b

What do you think?

graphium/trainer/predictor.py
WenkelF: Removed in a3d471540bc3f69cc0af2346ab0e66141b65311b as explained below

graphium/trainer/predictor.py
WenkelF: Good idea, this is implemented in a3d471540bc3f69cc0af2346ab0e66141b65311b

tests/dummy-pretrained-model-gpu.ckpt
DomInvivo: Yes, you can remove the gpu model.

graphium/cli/finetune.py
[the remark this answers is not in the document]
cwognum: I checked and it can be used with any of the training configs we had (the logic is the same as for `main_run_multitask.py`). Fine-tuning for now only works for ADMET, but as I commented above (and documented in the code), that can be easily extended. I will thus keep things named as is.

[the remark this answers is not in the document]
cwognum: My bad! I was looking at the wrong file! Blame the jetlag! 😉 

I still think we can keep the filename the same, since the CLI method is already very specifically named and it is easy to imagine that this will be extended in the future. This also provides contributors with a clear location to add finetuning CLIs.

WenkelF:
@cwognum the bug in the finetuning training is fixed here febdf2dcf21885e0c39e02d7caf64833e4e969d1

I missed a deepcopy operation when defining the datahashes for the TDC datasets. We include the first 5 rows of the df when generating the hash and the bug reduced the datasets to those 5 rows (molecules) as well.

Make sure to remove the TDC datasets from datacache. You can use `datamodule._path_to_load_from_file("train)` to get "train_[data_hash]" and then remove all data hashes (e.g., "train_[data_hash]", "test_[data_hash]", etc. that end with the same sequence.

WenkelF:
@DomInvivo I added some final improvements:
1. Made `modify_cfg_for_finetuning` function fully agnostic to pretrained model
2. Dropped dummy-pretrained-model-gpu and added `map_location` as an argument to  `load_pretrained_models `function
3. Added the option to keep modules after finetuning module

(3.) makes it much easier to finetune from modules other than the task_heads without manually re-defining the downstream network. When finetuning from task_heads, it is not needed.

s-maddrellmander:
@WenkelF - try merging from the main branch, I've made a change to the IPU test CLI that should take into account the changes made in this PR. If that doesn't work let me know.

DomInvivo (approved):

PR #418, opened 2023-07-24
Test ipu wrapper

callumm-graphcore:
## Changelogs

---

_Checklist:_

---

### DIFF
--- a/tests/test_ipu_wrapper.py
+++ b/tests/test_ipu_wrapper.py
@@ -3,7 +3,7 @@ import unittest as ut
 import torch
 from torch_geometric.data import Batch, Data
 
-from graphium.ipu_wrapper import PyGArgsParser
+from graphium.ipu.ipu_wrapper import PyGArgsParser
 
 def _make_fake_graph(num_nodes, num_edges, hidden_dim=16):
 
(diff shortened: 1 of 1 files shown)


DomInvivo (approved):

tests/test_ipu_wrapper.py
```diff
@@ -0,0 +1,40 @@
+import unittest as ut
+
+import torch
+from torch_geometric.data import Batch, Data
+
+from graphium.ipu.ipu_wrapper import PyGArgsParser
```
s-maddrellmander, line 6:
```suggestion
 # Run this test only if poptorch is available
            try:
                import poptorch
                from graphium.ipu.ipu_wrapper import PyGArgsParser
            except Exception as e:
                warn(f"Skipping this test because poptorch is not available.\n{e}")
                return

```
Try something like this - not quite sure, but we need to make sure the IPU files aren't imported when in the regular env. 

tests/test_ipu_wrapper.py:6: in <module>
    from graphium.ipu.ipu_wrapper import PyGArgsParser
graphium/ipu/ipu_wrapper.py:20: in <module>
    poptorch = import_poptorch()
graphium/ipu/ipu_utils.py:30: in import_poptorch
    raise e
graphium/ipu/ipu_utils.py:25: in import_poptorch
    import poptorch
E   ModuleNotFoundError: No module named 'poptorch'

s-maddrellmander: Tabs are a bit all over the apace on my suggestion there.

s-maddrellmander (changes requested):
Need to make this compatible with the regular env vs pop torch env.

callumm-graphcore:
Should we maybe add a decorator for tests that indicates they should only be run if IPU HW/SW is available?
________________________________

s-maddrellmander:
> Should we maybe add a decorator for tests that indicates they should only be run if IPU HW/SW is available?

We've got a pytest mark
@pytest.mark.ipu

Which will make sure the correct tests are picked up when we run in different conditions.

But I'm not certain about a decorator yet as we have a few different ways of setting up the tests for IPUs depending on what we need to do. But the mark should suffice for running the test.

The issue with the imports is they get imported on reading the file before getting to the test.

DomInvivo:
@callumm-graphcore any ETA for that PR?

DomInvivo:
@callumm-graphcore what's the status of this PR? 
I will close it and you can re-open it if it's still valid. But my understanding is that there's plenty of tests already added in other PR's, and that this is just a duplicate.

Issue #431, opened 2023-08-10
Issues for loading from RAM instead of from Disk


DomInvivo:
- [x] The `DatasetSubsampler` class fails if `processed_graph_data_path` is None, due to this line: https://github.com/datamol-io/graphium/blob/10fe04b2a7983523d9829ac37c4bfa3d8f7391c8/graphium/data/sampler.py#L28

- [x] `cache_data_path` should be depreciated in favor of `processed_graph_data_path`: https://github.com/datamol-io/graphium/blob/10fe04b2a7983523d9829ac37c4bfa3d8f7391c8/graphium/data/datamodule.py#L773

- [x] `load_from_file` should be its own parameter. Depending on `processed_graph_data_path` means that either:
  - We can cache the data and dataloading from disk
  - We cannot cache and we can do dataloading from RAM
  - We need the option of caching while doing dataloading from RAM, and this will be enabled by a new `load_from_file` parameter in the Datamodule

- [x] Why is `normalize_label` only applied when dataloading is from RAM? Where is it applied on disk? https://github.com/datamol-io/graphium/blob/10fe04b2a7983523d9829ac37c4bfa3d8f7391c8/graphium/data/datamodule.py#L1300

WenkelF:
> Why is `normalize_label` only applied when dataloading is from RAM? Where is it applied on disk?

This is fine because `_save_data_to_files()` internally calls `_make_multitask_dataset(..., load_from_file=False)`, i.e., label normalization is already applied to datasets loaded from disk.

PR #432, opened 2023-08-10
Caching logic improvement

DomInvivo:
## Changelogs

This is a draft PR looking to change the logic of how caching is done. The PR is motivated in #431 . I'll wait for comments and suggestions before pursuing this.

- Removed the `cache_data_path` option. We don't want to load 
- Added the `dataloading_from` option to select whether to load from Disk or RAM

---

_Checklist:_

- [x] _Was this PR discussed in an issue? _ Yes in #431 

---


graphium/data/datamodule.py
```diff
@@ -1160,13 +1170,9 @@ def has_atoms_after_h_removal(smiles):
             self.single_task_datasets, self.task_train_indices, self.task_val_indices, self.task_test_indices
         )
 
-        if self.load_from_file:
+        if self.processed_graph_data_path is not None:
             self._save_data_to_files()
```
WenkelF, line 1174: @DomInvivo not sure this is relevant, but we always cache data if `processed_graph_data_path` is provided.

Are there situations, one wants to read from disk during running the pipeline (e.g., if model doesn't fit in ram otherwise) but discard the data afterwards to not block disk space?

DomInvivo: If you read from disk, then the data is already there. Deleting it would mean to delete the data from a previous run

graphium/data/datamodule.py
```diff
@@ -1185,7 +1191,7 @@ def setup(
         labels_size = {}
         labels_dtype = {}
         if stage == "fit" or stage is None:
-            if self.load_from_file:
+            if self.processed_graph_data_path is not None:
```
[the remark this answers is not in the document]
DomInvivo: You're right. Just fixed it

graphium/data/datamodule.py
```diff
@@ -1286,8 +1291,8 @@ def _make_multitask_dataset(
             progress=self.featurization_progress,
             about=about,
             save_smiles_and_ids=save_smiles_and_ids,
-            data_path=self._path_to_load_from_file(stage) if load_from_file else None,
-            load_from_file=load_from_file,
+            data_path=self._path_to_load_from_file(stage) if processed_graph_data_path else None,
+            processed_graph_data_path=processed_graph_data_path,
```
WenkelF, line 1295:
```suggestion
            dataloading_from=self.dataloading_from
```
We don't need `processed_graph_data_path` here but rather `data_path` as it is otherwise hardcoded to "ram" in the MultitaskDataset

graphium/data/datamodule.py
WenkelF: We need to check if the data is already cached. Otherwise, [this funciton](https://github.com/datamol-io/graphium/blob/7fa23b6fcc59dd07c8f2459706d432e97f6d0f21/graphium/data/datamodule.py#L1435-L1436) will throw an error.

WenkelF:
@DomInvivo I tested the caching branch and at closer inspection found some bugs that were not evident from the changelogs:

The unit tests currently fail because we need to add `from os import PathLike as Path` to the imports in the `datamodule` or directly use `os.PathLike` instead.

After this is fixed, there is an issue because the `MultitaskDataset` is defined as follows:
https://github.com/datamol-io/graphium/blob/7fa23b6fcc59dd07c8f2459706d432e97f6d0f21/graphium/data/dataset.py#L136-L151

However, in the `datamodule` we call it like this:
https://github.com/datamol-io/graphium/blob/7fa23b6fcc59dd07c8f2459706d432e97f6d0f21/graphium/data/datamodule.py#L1286-L1297 

It seems `dataloading_from` is hardcoded to `“ram”` and we pass the unsupported `processed_graph_data_path` as an argument. Instead, we need to pass `dataloading_from` here (see proposed change above).

With this change, it runs for `“disk”`. **IF** we don’t set `processed_graph_data_path`, it also runs for `“ram”`. If we set `“ram”` and `processed_graph_data_path`, we need to check if data is already cached, otherwise we get an error when trying to create already existing files (see comment above).

WenkelF:
@DomInvivo here are the main updates:
- Simplifying logic to prepare data and do dataloading from `DISK/RAM`
- Allow cached data (on `DISK`) to be loaded into `RAM` for dataloading
- Speeding up transfer from `DISK` to `RAM` via parallelization
- Adding option to prepare data in advance to `CLI` via `graphium-prepare-data` (make sure to re-install dependencies before using it)
- Updating `README` to explain how and why to prepare data in advance

graphium/cli/prepare_data.py
```diff
@@ -0,0 +1,40 @@
...
+    The main (pre-)training and fine-tuning loop.
+    """
+
+    cfg = OmegaConf.to_container(cfg, resolve=True)
+
+    st = timeit.default_timer()
+
+    ## == Instantiate all required objects from their respective configs ==
+    # Accelerator
+    cfg, accelerator_type = load_accelerator(cfg)
```
DomInvivo, line 29: You can enforce the `accelerator_type` to CPU when preparing the data. Don't need this line

graphium/cli/prepare_data.py
```diff
@@ -0,0 +1,40 @@
...
+
+    ## Data-module
+    datamodule = load_datamodule(cfg, accelerator_type)
```
DomInvivo, line 32: Before going here, I would raise an error if `cfg["datamodule"]["args"]["processed_graph_data_path"]` is not present or None. As we don't want to be preparing the data without storing them to disk, it would be a waste of compute

DomInvivo (commented):
Minor comments. Good work :)

WenkelF:
@DomInvivo some finishing touches for the PR:
- Fixed unit tests for test_datamodule
- Adding warning message `save_smiles_and_ids` argument in `setup()` is superseeded by `prepare_data()`
- Temporary hardcoding no parallelization when transerring from DISK to RAM

WenkelF:
@DomInvivo should be ready to merge

PR #433, opened 2023-08-12
Adding unit test for finetuning from gnn

WenkelF:
@DomInvivo the new unit test covers the following additional features:
- Finetuning from gnn (formerly only task head)
- Keeping & modifying modules of pretrained model after finetuning module
- Unfreezing additional pretrained depth


WenkelF:
It runs about 2 minutes

WenkelF:
@DomInvivo I reduced the runtime of the test to 61 seconds by reducing the number of epochs of finetuning from 3 to 2 5ef06f2fafe52c2ced3bf3db41498dca4ba2ecf3

WenkelF:
@DomInvivo it now runs in <10 seconds after setting `sample_size=10`

PR #434, opened 2023-08-14
Disable wandb by default + complete integration tests

s-maddrellmander:
## Changelogs

- Making sure the default behaviour is sensible, i.e. users aren't having to bodge a lack of wandb accounts + login details / project access etc. 
- Adding tests for the new mode - ipu and cpu, complete integration tests for training + fine tuning

---


tests/test_ipu_training.py
cwognum, line 1: This feels like a lot of code duplication compared to the "normal" training / finetuning script! Is that needed? Reducing duplication reduces confusion and makes it easier to maintain the code base.

graphium/cli/train_finetune.py
```diff
@@ -44,6 +44,14 @@ def run_training_finetuning(cfg: DictConfig) -> None:
 
     st = timeit.default_timer()
 
+    # Disable wandb if the user is not logged in.
+    if wandb.login() is False:
```
cwognum, line 48: Nice! I like this! Could we merge it with the logic I introduced (i.e. checking if the `constants.wandb` key exists)? Or do both approaches complement each other?

s-maddrellmander: I think this is useful for certain environments where the user may not be logged in but the config is set, especially where we are both using configs for experiments + public release. But no strong feelings about exactly the mode to check, but this seemed simple enough.

tests/test_ipu_training.py
s-maddrellmander: Yeah, still WIP, and wanting to get the hydra integration working for CPU/IPU in the test env first within a single test. 
And hadn't decided between cutting the script down, or reusing the main code. But good point. :)

cwognum: Great, thanks!

graphium/cli/train_finetune.py
s-maddrellmander: Yeah I think you're right! I'll make it a two part with an `elif`.

tests/test_ipu_training.py
```diff
@@ -0,0 +1,41 @@
...
+
+@pytest.mark.parametrize("acc_type", ["cpu", "ipu"])
+@pytest.mark.ipu
+def test_cli(acc_type) -> None:
+    """
+    The main CLI endpoint for training and fine-tuning Graphium models.
+    """
+    with hydra.initialize(version_base=None, config_path="../expts/hydra-configs"):
+        # config is relative to a module
+        cfg = hydra.compose(config_name="test", overrides=[f"accelerator={acc_type}"])
```
cwognum, line 36: Hey @s-maddrellmander, one more question I wanted to discuss: 

Do we need as many config files? 

Hydra supports a very elaborate [override syntax](https://hydra.cc/docs/advanced/override_grammar/basic/). I believe it will benefit the clarity and maintainability of the config system, if we actively try to minimize the number of configs laying around. Hydra automatically saves the full config to the [working/output directory](https://hydra.cc/docs/tutorials/basic/running_your_app/working_directory/#changing-or-disabling-hydras-output-subdir), so you can still easily save experiment-specific configs for reproducibility.

As for this specific case: I don't know for sure (please let me know what you think!), but I can imagine that we don't need 9 new configs, because we only want to override a few keys in the config (e.g. `constants.max_epochs` or `sample_size`). I now expect there to be a lot of repetition with the existing training configs. A simple example, instead of creating `expts/hydra-configs/test.yaml`, you could also do: 

```python
cfg = hydra.compose(config_name="test", overrides=[
    f"accelerator={acc_type}",
    "architecture=test",
    "task=test",
    "training=test",
])
```

I would prefer to have a long overrides list over many different new files. 

Let me know what you think!

s-maddrellmander: That's fair, still getting my head round all the hydra functionality, I'll look at adapting the existing configs via the override as much as possible. That'll be neater than having so many additional files to maintain.

cwognum: This is a lot better! Thank you for taking the time 🙏

graphium/cli/train_finetune.py
```diff
@@ -44,8 +44,23 @@ def run_training_finetuning(cfg: DictConfig) -> None:
...
+        logger.info(
+            "Not logged in to wandb - disabling wandb logging.\n"
+            + "To enable wandb, run `wandb login` from the command line."
+        )
+        wandb.init(mode="disabled")
+
```
cwognum, line 54: This piece of code is now a duplicate of L57 to L62, right? We should be able to delete it.  

```suggestion
```

cwognum, line 54: This piece of code is now a duplicate of L57 to L62, right? We should be able to delete it.  

```suggestion
```

s-maddrellmander: Whoops! Good catch!

graphium/cli/train_finetune.py
```diff
@@ -44,8 +44,15 @@ def run_training_finetuning(cfg: DictConfig) -> None:
...
+    if wandb.login() is False:
+        logger.info(
+            "Not logged in to wandb - disabling wandb logging.\n"
+            + "To enable wandb, run `wandb login` from the command line."
+        )
+        wandb.init(mode="disabled")
+    elif wandb_cfg is not None:
         wandb.init(
             entity=wandb_cfg["entity"],
             project=wandb_cfg["project"],
```
DomInvivo, line 58: This will allow to pass more than just the entity or the project.
```suggestion
    elif wandb_cfg is not None:
        wandb.init(
            config=cfg,
            **wandb_cfg
```

tests/test_ipu_training.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.mark.ipu
+def test_cli(acc_type) -> None:
+    """
+    The main CLI endpoint for training and fine-tuning Graphium models.
+    """
+    with hydra.initialize(version_base=None, config_path="../expts/hydra-configs"):
+        # config is relative to a module
+        cfg = hydra.compose(
+            config_name="main",
+            overrides=[
```
DomInvivo, line 18: Also override 
```yaml
trainer.trainer.check_val_every_n_epoch = 1
trainer.trainer.precision = 32, # perhaps you can make this 32 for CPU and 16 for IPU
```

DomInvivo (approved):
Looks good! Minor changes only

PR #441, opened 2023-08-22
WIP: Various improvements to the CLI endpoints

cwognum:
## Changelogs

- Added fine-tuning through [Optuna](https://optuna.org/), using the [Optuna Sweeper Plugin](https://hydra.cc/docs/plugins/optuna_sweeper/) for Hydra. 
- Migrated the CLI from [Click](https://click.palletsprojects.com/en/8.1.x/) to [Typer](https://typer.tiangolo.com/).
- Made it possible to specify a pre-trained model with a `.ckpt` path, instead of always relying on a manually defined [mapping](https://github.com/datamol-io/graphium/blob/main/graphium/utils/spaces.py#L128-L130) in the `spaces.py`.
- Added some utilities for saving essential results in case there is no logger.
- Removed the `graphium-prepare-data` command in favor of a new subcommand at: `graphium data prepare`. 
- Added `graphium finetune fingerprint` to easily extract fingerprints from a model.
- _(WIP: Added a notebook to investigate the correlation between pre-training and fine-tuning performance.)_

---

_Checklist:_

- [ ] ~_Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~
- [X] _Write concise and explanatory changelogs above._
- [X] _If possible, assign one of the following labels to the PR: `feature`, `fix` or `test` (or ask a maintainer to do it for you)._

---


notebooks/compare-pretraining-finetuning-performance.ipynb
DomInvivo, line 1: Here you plot the loss and the fine-tuning performance separately in cells [7] and [8]. However, it's difficult to see if there's any correlation at all.

Ideally, a scatter plot with 
- x axis : the loss (or any other metric) on the test set of the ToyMix
- y axis: the loss (or any other metric) on the test set of the fine-tuning task
- Add a correlation between the axes similar to [seaborn jointplot](https://seaborn.pydata.org/examples/regression_marginals.html)
- Ideally a [multiplot ](https://seaborn.pydata.org/_images/axis_grids_32_0.png)with many metrics and many datasets

README.md
```diff
@@ -105,7 +105,7 @@ However, when working with larger datasets, it is recommended to perform data pr
 The following command-line will prepare the data and cache it, then use it to train a model.
 ```bash
 # First prepare the data and cache it in `path_to_cached_data`
-graphium-prepare-data datamodule.args.processed_graph_data_path=[path_to_cached_data]
+graphium data prepare ++datamodule.args.processed_graph_data_path=[path_to_cached_data]
```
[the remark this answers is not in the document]
cwognum: `graphium-train` is the hydra endpoint - specifically meant for training and finetuning. `graphium` is the more general CLI endpoint, organized with various sub commands. Ideally I would've also made the training and finetuning accessible under the `graphium` command, for example under the `graphium train` subcommand, but as I tried to explain in https://github.com/datamol-io/graphium/pull/432#issuecomment-1690082766, the hydra CLI cannot be a subcommand of another CLI, which is why we had to [define](https://github.com/datamol-io/graphium/blob/main/pyproject.toml#L65-L68) a second command. Since the command cannot have any spaces, I think this is a reasonable name for it.

docs/cli_references.md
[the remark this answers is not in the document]
cwognum: Thinking about this some more, there is a case to be made that it simplifies indexing and searching the docs as a CLI becomes complex, but I don't think that is worth it for Graphium as of now as the CLI is very simple.

expts/hydra-configs/hparam_search/optuna.yaml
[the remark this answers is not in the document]
cwognum: Fair question!

Optuna has more advanced features. For example, W&B sweeps support only [three search strategies](https://docs.wandb.ai/guides/sweeps/define-sweep-configuration#configuration-keys), whereas Optuna supports [many more](https://optuna.readthedocs.io/en/stable/reference/samplers/index.html). Another example would be that Optuna supports [MPO](https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/002_multi_objective.html). Furthermore, I like that Optuna is free to us, open source and has zero barriers to start using it (e.g. you don't even need an account). I think that fits an open-source library like Graphium. The use of W&B should never be needed. 

Of course, nothing prevents you from still logging all results to W&B with Pytorch Lightning's or [Optuna's own](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.WeightsAndBiasesCallback.html#optuna.integration.WeightsAndBiasesCallback) callback to use all of their nice dashboards and visualization. 

On a personal note: My W&B account has exceeded all limits on the free subscription plan, so I cannot really use it anymore.

graphium/cli/train_finetune.py
```diff
@@ -38,6 +51,18 @@ def run_training_finetuning(cfg: DictConfig) -> None:
...
     cfg = OmegaConf.to_container(cfg, resolve=True)
 
+    dst_dir = cfg["constants"].get("results_dir")
+    hydra_cfg = HydraConfig.get()
+    output_dir = hydra_cfg["runtime"]["output_dir"]
+
+    if dst_dir is not None and fs.exists(dst_dir) and len(fs.get_mapper(dst_dir).fs.ls(dst_dir)) > 0:
+        logger.warning(
+            "The destination directory is not empty. "
+            "If files already exist, this would lead to a crash at the end of training."
```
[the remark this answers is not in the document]
cwognum: Because you're not sure it will crash. The exact filenames that will be saved to the destination directory depend on the configuration and there is no easy way to figure them out. Similar to how the PTL `ModelCheckpoint` callback also just [warns](https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/callbacks/model_checkpoint.html#ModelCheckpoint.setup) you when the directory is not empty.

graphium/cli/train_finetune.py
```diff
@@ -105,6 +130,12 @@ def run_training_finetuning(cfg: DictConfig) -> None:
     with SafeRun(name="TRAINING", raise_error=cfg["constants"]["raise_train_error"], verbose=True):
         trainer.fit(model=predictor, datamodule=datamodule)
 
+    # Save validation metrics - Base utility in case someone doesn't use a logger.
+    results = trainer.callback_metrics
+    results = {k: v.item() if torch.is_tensor(v) else v for k, v in results.items()}
+    with fsspec.open(fs.join(output_dir, "val_results.yaml"), "w") as f:
```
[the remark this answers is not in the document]
cwognum: In case W&B is used, all these results are already logged to W&B, no? 

But yes, you could configure the `output_dir` through the config ([ref](https://hydra.cc/docs/configure_hydra/workdir/)) and have it be in the W&B dir.

notebooks/compare-pretraining-finetuning-performance.ipynb
cwognum: As said during our meeting; These plots are WIP and I might complete redo them once we have actual results. At that time, I will also motivate the plots I came up with. 

> Here you plot the loss and the fine-tuning performance separately in cells [7] and [8]. However, it's difficult to see if there's any correlation at all.

No, that's incorrect. I plot them jointly. See the axes labels. 

> Ideally, a scatter plot

Yeah, was planning of adding it, but I don't think it's the most informative. 

For each pre-trained model (e.g. on LargeMix), you will have a _single_ score (i.e. the test loss) that you're trying to correlate with the metrics of _many_ fine-tuned models. That will lead to lines of dots.

graphium/cli/finetune_utils.py
```diff
@@ -70,9 +63,70 @@ def benchmark_tdc_admet_cli(save_dir, wandb, name, inclusive_filter):
...
+    save_dir = cfg["constants"].get("results_dir", HydraConfig.get()["runtime"]["output_dir"])
     fs.mkdir(save_dir, exist_ok=True)
     path = fs.join(save_dir, "results.yaml")
     logger.info(f"Saving results to {path}")
 
     with fsspec.open(path, "w") as f:
         yaml.dump(results, f)
+
+
+@finetune_app.command(name="fp")
```
DomInvivo, line 77: is `fp` well known? why not `fingerprint` or `feature_vector`?

graphium/cli/finetune_utils.py
```diff
@@ -70,9 +63,70 @@ def benchmark_tdc_admet_cli(save_dir, wandb, name, inclusive_filter):
...
+        pretrained_model,
+        device=accelerator_type,
+    ).model
+
+    ## == Fingerprinter
+
+    fps = []
+    with Fingerprinter(network=model, fingerprint_spec=fingerprint_layer_spec) as fp:
+        for batch in tqdm.tqdm(datamodule.predict_dataloader(), desc="Getting fingerprints"):
+            # TODO (cwognum): This is a hack to make sure the features are float32 (instead of float16)
```
DomInvivo, line 117: What hack? I don't see a hack?

graphium/finetuning/fingerprinting.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+
+        # Run the batch through the model.
+        with torch.inference_mode():
+            self.network(batch["features"])
+
+        readout_list = []
+        for module_name, layers in self._spec.items():
+            readout_list.extend(
+                [self.network._module_map[module_name]._readout_cache[layer] for layer in layers]
```
DomInvivo, line 121: Add `.detach().cpu()` to kill the compute graph (although might not be there in inference mode), and move the data to CPU, otherwise large datasets could overflow the GPU.

notebooks/compare-pretraining-finetuning-performance.ipynb
```diff
@@ -0,0 +1,326 @@
...
+   ],
+   "source": [
+    "sns.heatmap(data, annot=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The End."
```
DomInvivo, line 300: Love this <3

graphium/cli/finetune_utils.py
cwognum: I think it is well known, but I'm in favor of making it more explicit! 👍

graphium/cli/finetune_utils.py
cwognum: I was referring to: 

```python
if isinstance(v, torch.Tensor) and torch.is_floating_point(v):
    batch["features"][k] = v.float()
```

This feels too hard-coded to me. Is it reasonable to always use this? E.g. does the model always have the `float32` dtype?

graphium/finetuning/fingerprinting.py
cwognum: `.detach()` won't be needed, but we should indeed move it to the CPU!

notebooks/compare-pretraining-finetuning-performance.ipynb
cwognum: FINALLY someone that notices! ❤️ 

I'm trying to make it my signature! See e.g. [MOOD](https://github.com/valence-labs/mood-experiments/tree/main/notebooks). 😅

graphium/cli/finetune_utils.py
DomInvivo: You just have to make sure that, in the trainer, you use `precision=32`. Or is the problem here that you're not using the trainer, and it's unclear what's the dtype input?

I think that any dtype works due to this, called at the beginning of the `forward` https://github.com/datamol-io/graphium/blob/f96c1b60b4d86e2fde0f3660b2916c14de388749/graphium/trainer/predictor.py#L198

graphium/cli/finetune_utils.py
DomInvivo: More explicit please :)

notebooks/compare-pretraining-finetuning-performance.ipynb
DomInvivo: no, not a single score, many scores, e.g. the loss on PCBA, the pearson on PCQM

graphium/cli/train_finetune.py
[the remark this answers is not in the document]
DomInvivo: yeah, but you'll get the ability of downloading the yaml directly from wandb.

docs/cli_references.md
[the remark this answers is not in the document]
DomInvivo: I personally prefer the website option to the CLI help. And as you mentioned, simplifies the searching + indexing, and makes it more explicit that there's an advanced CLI associated to the library (it's not the case for most libraries).

I want to avoid duplicating the maintenance work, so is there a way to automatically link the docs page to the CLI docs directly?

README.md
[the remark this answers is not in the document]
DomInvivo: Okok I see

docs/cli_references.md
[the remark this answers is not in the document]
cwognum: For `typer`, it seems we can [automatically generate docs and save them to a README file](https://typer.tiangolo.com/typer-cli/#generate-docs) with the [`Typer CLI` extension](https://typer.tiangolo.com/typer-cli/). Maybe we could make that part of the doc deployment CICD? As for hydra, as of now there seems to be no straight-forward way of doing this, e.g. see https://github.com/facebookresearch/hydra/issues/633.

Besides the auto-generated docs, I think it makes sense to either way also include a short tutorial / doc page with an explanation. I will propose something!

graphium/cli/finetune_utils.py
cwognum: Yeah, indeed, felt like a hack because I wasn't using the `predictor`. Changed this based on your comment! Thanks! 🙏

DomInvivo (approved):
Only a comment about the CLI for fingerprinting. Seems to miss the ability to chose a dataset?? Or does that come from the config?

docs/cli/graphium.md
```diff
@@ -0,0 +1,156 @@
...
+
+* `OVERRIDES...`: [required]
+
+**Options**:
+
+* `--name TEXT`
+* `--inclusive-filter / --no-inclusive-filter`: [default: inclusive-filter]
+* `--help`: Show this message and exit.
+
+### `graphium finetune fingerprint`
```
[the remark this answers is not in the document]
cwognum: Comes from the config! It uses the `predict_dataloader()` of the datamodule, which in Graphium has been implemented to fallback to the `test_dataloader` if not explicitly specified.

PR #449, opened 2023-08-28
Various updates to graphium

WenkelF:
## Changelogs

1. Further improvement of caching logic (only calling `datamodule._make_multitask_dataset()` once)
2. Integrated LargeMix into `hydra`
3. Integrated single datasets (LargeMix) into `hydra`
4. Computing val/test metrics on `cpu` to save `gpu` memory
5. Implemented testing using a model checkpoint

## Remarks
**Regarding(3.):** Single dataset runs are specific to NeurIPS experiment, where we use the same model as for LargeMix, i.e., run them like follows:
```
graphium-train \
    architecture=largemix \
    tasks=l1000_mcf7 \
    training=largemix \
```
**Regarding (5):** Needs to be properly changed to `typer` following comment by @cwognum  (incl. changing documentation) once branch has been updated with `main`.

---

_Checklist:_

- ~~[ ] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~~

---


expts/hydra-configs/architecture/largemix.yaml
DomInvivo, line 1: All those yaml files seem like duplicate from the files in the `neurips2023` folder. What do you think? Is there a reason to have them separately?

graphium/config/_loader.py
```diff
@@ -589,15 +589,15 @@ def get_checkpoint_path(config: Union[omegaconf.DictConfig, Dict[str, Any]]) ->
...
     cfg_trainer = config["trainer"]
 
-    path = config.get("ckpt_name_for_testing", "last.ckpt")
-    if path in GRAPHIUM_PRETRAINED_MODELS_DICT or fs.exists(path):
-        return path
-
     if "model_checkpoint" in cfg_trainer.keys():
-        dirpath = cfg_trainer["model_checkpoint"]["dirpath"]
-        path = fs.join(dirpath, path)
+        dirpath = cfg_trainer["model_checkpoint"]["dirpath"] + str(cfg_trainer["seed"]) + "/"
```
DomInvivo, line 593: Can you create a unique identifier that you can associate with a specific run? Perhaps take the identifier from Wandb?

tests/test_datamodule.py
```diff
@@ -55,8 +55,7 @@ def test_ogb_datamodule(self):
             rm(TEMP_CACHE_DATA_PATH, recursive=True)
 
         # Reset the datamodule
-        ds._data_is_prepared = False
-        ds._data_is_cached = False
+        ds = GraphOGBDataModule(task_specific_args, **dm_args)
```
DomInvivo, line 58: What's the purpose of this??

graphium/config/_loader.py
cwognum: Unless you want to add an ID that is only programmatically accessible, I would leave it to the config and not hard-code the seed in the path since it is pretty subjective where you want to save the models. Note that setting the dirpath to include the seed can currently also be achieved by changing the config - either directly or through the CLI.

DomInvivo: We have the issue that we want identifiable checkpoints, and non- overwriting checkpoints. How do you suggest we move forward?

cwognum: I understand the use-case! 

As I said before, I suggest using the config. I would set a sensible default, but then leave it up to the user. 

E.g. the line that is hard coded here can already be configured with [`hydra`s override syntax](https://hydra.cc/docs/advanced/override_grammar/basic/), using [OmegaConf's interpolation](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#variable-interpolation): 
```sh
graphium-train ... "trainer.model_checkpoint.dirpath='/path/to/checkpoints/${constants.seed}'"
```

In case of a sweep, you might want to use something more specific, e.g.: 
```sh
graphium-train -m ... "trainer.model_checkpoint.dirpath='/path/to/checkpoints/${now:%Y-%m-%d_%H-%M-%S}'"
```

To prevent most users from having to set a value themselves, you could include a sensible default in the config files. Current defaults (e.g. `models_checkpoints/neurips2023-small-gcn/`) are too specific and not consistent. I would e.g. save to the [current working dir](https://hydra.cc/docs/configure_hydra/workdir/) by default: 

```yaml
trainer:
  model_checkpoint:
    dirpath: ${hydra.run.dir}
```

DomInvivo: Thanks for the suggestion. I'm still getting used to hydra and didn't know you could put the date-time directly in the Yaml! What do you think of [this commit](https://github.com/datamol-io/graphium/pull/449/commits/da31797946fe2f094c355aca452f0df5d872fcbe)? Making it the default in all the `hydra-configs`

DomInvivo: Or do you mean that, by default, the `{hydra.run.dir} includes the date-time?

WenkelF: I agree that we have to leave the checkpoint path configuration somewhat up to the user.

However, the seed is an important culprit leading to overwriting (or unclear identity) of model checkpoints. I post-pended it to the `dirpath `because we also used to have the following in `load_trainer`:
```
if "model_checkpoint" in cfg_trainer.keys():
    cfg_trainer["model_checkpoint"]["dirpath"] += str(cfg_trainer["seed"]) + "/"
    callbacks.append(ModelCheckpoint(**cfg_trainer["model_checkpoint"]))
```

WenkelF: I am using `graphium-train ... "trainer.model_checkpoint.dirpath='/path/to/checkpoints/${constants.seed}'"` as proposed by Cas to reasonbly structure checkpoints

tests/test_datamodule.py
WenkelF: This is properly resetting the `datamodule`. The old code didn't reset `self.train_ds`, etc., which could lead to wrong testing as, e.g., `setup` checks `if self.train_ds is None`

WenkelF: But probalby we can remove this function and everything related to testing as @cwognum used `get_checkpoint_path` it in https://github.com/datamol-io/graphium/pull/441 where he has made some changes to testing :)

maciej-sypetkowski:
> @maciej-sypetkowski , do you think that moving the val/test metrics computations to CPU would significantly hurt parallelization?

I think it should be fine, however the predictions are still kept on gpus if I look correctly (the casting is only in `_general_epoch_end`).

> We are currently constrained as we don't use the torchmetrics ability to update stats due to their inability to handle NaNs.

It can be done in torchmetrics, you just need to implement a wrapper for ignoring nans. We'll probably want to do this when we add better multi-device support.

DomInvivo:
@maciej-sypetkowski 
> It can be done in torchmetrics, you just need to implement a wrapper for ignoring nans. We'll probably want to do this when we add better multi-device support.

We already have the `MetricWrapper`, but it works on the metric level, not on the `update` and `compute` level. I guess that's a good place to start looking in the future.

graphium/config/_loader.py
DomInvivo: Why not use the full date and time? Much better than the random seed. And avoids collision when doing sweeps with pre-selected seeds. And also interpretable.

WenkelF: Yes, I agree. It was necessary at the time the branch was created because the seed was hard-coded into the `dirpath` back then... I think it is good as of now using the date :)

expts/hydra-configs/architecture/largemix.yaml
WenkelF: I curated the configs to make LargeMix configs easy to use within our hydra logic. In particular, the modularity with the added files will facility finetuning on pretrained models on LargeMix. As far as I see, neurips2023 configs can only be used for training on LargeMix, but one needs to write new configs for, e.g., finetuning or preparing data.

graphium/config/_loader.py
WenkelF: Changed function discussed here to its current state in the main branch. The function was outdated in this branch because we added it to https://github.com/datamol-io/graphium/pull/441/files where it was modified to accommodate further changes :)

WenkelF:
@DomInvivo I think we can merge. I suggest opening a new branch for the optimizations discussed with @maciej-sypetkowski

PR #451, opened 2023-08-30
Node inconsistent hydrogens number

DomInvivo:
## Changelogs

- The previous code was allowing inconsistent number of labels per molecule to go through, leading to errors if there are mistakes in the dataset or the choice of `label_cols`. For instance, if we have the cols `node_feat1` of size 10, and `node_feat2` of size 12, but they're both for the same molecule, then the code should either crash, or filter out the inconsistencies.

@engmubarak48 and @luis-mueller , I'm not sure I understand all the logic implemented for the padding, so I'd like your opinion on this PR.

---

_Checklist:_

---

### DIFF
--- a/graphium/data/multilevel_utils.py
+++ b/graphium/data/multilevel_utils.py
@@ -40,14 +40,12 @@ def extract_labels(df: pd.DataFrame, task_level: str, label_cols: List[str]):
     def merge_columns(data: pd.Series):
         data_name = data.name
         data = data.to_list()
-        missing_label_idx = [i for i, d in enumerate(data) if not isinstance(d, np.ndarray) and math.isnan(d)]
-        data = [np.array([np.nan]) if i in missing_label_idx else d for i, d in enumerate(data)]
-
+        data = [np.array([np.nan]) if not isinstance(d, np.ndarray) and math.isnan(d) else d for d in data]
         # Ensure that all shapes are either the same, or 1
         sizes = [d.shape[0] for d in data]
         maxsize = max(sizes)
         for i, size in enumerate(sizes):
-            if i in missing_label_idx:
+            if size == 1 and np.isnan(data[i]).all():
                 continue
             if size != maxsize:
                 raise ValueError(
--- a/graphium/data/multilevel_utils.py
+++ b/graphium/data/multilevel_utils.py
@@ -38,6 +38,7 @@ def extract_labels(df: pd.DataFrame, task_level: str, label_cols: List[str]):
         return data.apply(unpack)
 
     def merge_columns(data: pd.Series):
+        data_name = data.name
         data = data.to_list()
         missing_label_idx = [i for i, d in enumerate(data) if not isinstance(d, np.ndarray) and math.isnan(d)]
         data = [np.array([np.nan]) if i in missing_label_idx else d for i, d in enumerate(data)]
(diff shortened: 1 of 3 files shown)


graphium/data/multilevel_utils.py
```diff
@@ -39,6 +40,17 @@ def unpack_column(data: pd.Series):
     def merge_columns(data: pd.Series):
         data = data.to_list()
         data = [np.array([np.nan]) if not isinstance(d, np.ndarray) and math.isnan(d) else d for d in data]
+
+        # Ensure that all shapes are either the same, or 1
+        sizes = [d.shape[0] for d in data]
+        maxsize = max(sizes)
+        for size in sizes:
+            if (size != maxsize) and (size != 1):
```
luis-mueller, line 48: I assume the `size != 1` catches the nan-array created in line 42? It would probably be safer to specificially check for the nan-array. Let me think of a solution.

graphium/data/multilevel_utils.py
```diff
@@ -39,6 +40,17 @@ def unpack_column(data: pd.Series):
     def merge_columns(data: pd.Series):
         data = data.to_list()
         data = [np.array([np.nan]) if not isinstance(d, np.ndarray) and math.isnan(d) else d for d in data]
```
luis-mueller, line 42:
```suggestion
        missing_label_idx = [i for i, d in enumerate(data) if not isinstance(d, np.ndarray) and math.isnan(d)]
        data = [np.array([np.nan]) if i in missing_label_idx else d for i, d in enumerate(data)]
```

graphium/data/multilevel_utils.py
luis-mueller, line 48:
```suggestion
            if i in missing_label_idx:
                continue
            if (size != maxsize):
```

graphium/data/multilevel_utils.py
```diff
@@ -39,6 +40,17 @@ def unpack_column(data: pd.Series):
...
+
+        # Ensure that all shapes are either the same, or 1
+        sizes = [d.shape[0] for d in data]
+        maxsize = max(sizes)
+        for size in sizes:
```
luis-mueller, line 47:
```suggestion
        for i, size in enumerate(sizes):
```

luis-mueller (commented):
See my suggestion. I think your implementation would also allow a case where we e.g., have 10 node labels in one column and 1 node label in another column.

luis-mueller:
The `[nan]` was just a simple way to cast every column to a numpy array with 1 dim (columns could have simply a `nan` instead of an array). Let me fix the tests.

graphium/data/multilevel_utils.py
```diff
@@ -37,8 +38,21 @@ def unpack_column(data: pd.Series):
...
 
     def merge_columns(data: pd.Series):
+        data_name = data.name
         data = data.to_list()
         data = [np.array([np.nan]) if not isinstance(d, np.ndarray) and math.isnan(d) else d for d in data]
+        # Ensure that all shapes are either the same, or 1
+        sizes = [d.shape[0] for d in data]
+        maxsize = max(sizes)
+        for i, size in enumerate(sizes):
+            if size == 1 and np.isnan(data[i]).all():
```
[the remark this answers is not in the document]
DomInvivo: Thanks a lot!

PR #457, opened 2023-09-04
Pipeline integration + Virtual Nodes Edges bug fix

s-maddrellmander:
## Changelogs

- Stitching the accelerator kwargs into the model
- Found + fixed a bug that meant edges wouldn't work with the virtual nodes
- Added a function to recursively check for the edge out dim in the GNN layer kwargs - preserves previous functionality for simple layers and adds functionality for more complex layers.


kerstink-GC (approved):
Looks good

PR #472, opened 2023-09-26
Option to filter molecules post featurisation to limit moleucle size

s-maddrellmander:
## Changelogs

- The datasets tend to have a very long tail of large molecules. To efficiently make batches for training it's a good idea to limit the upper size of molecules. 
- This is done with a simple filter function that extends the null featurisation function.
- Simply add a line: `max_num_heavy_atoms: 100` under `data module.args` 

---


s-maddrellmander:
Screenshot 2023-09-26 at 12 31 01
See here the reduction of a few molecules in my small test

DomInvivo:
@s-maddrellmander , this is a duplicated code. Check the option `max_num_atoms` in the `featurization_kwargs`. Molecules that don't respect that threshold get featurized to None and filtered out

PR #477, opened 2023-10-16
Various updates

WenkelF:
## Changelogs

- Integrated resuming training from checkpoint when checkpoint path is provided in config under key `resume_from_checkpoint`
- Integrated testing from best checkpoint when using early stopping (useable by adding below information to config)
```
trainer:
  model_checkpoint:
    ...
    save_top_k: 1
    every_n_epochs: 1
    dirpath: [tbd]
    filename: best
  test_from_checkpoint: best.ckpt
```
- Allowing sampling fractions in `(0,1)` to scale _downscale_ dataset size
- Bug fixes when finetuning mup-scaled models
- Added functionalities for automated finetuning on TDC
  - [Baseline config](https://github.com/datamol-io/graphium/blob/updates_fred/expts/hydra-configs/finetuning/admet_baseline.yaml)
  - [Functionalities for easy configuration](https://github.com/datamol-io/graphium/blob/3895d239dc69c82352c159cf543502f4ff078c63/graphium/cli/finetune_utils.py#L139-L172)
- Added [code to infer & balance parameter count](https://github.com/datamol-io/graphium/blob/updates_fred/graphium/cli/parameters.py) for target model size

---

_Checklist:_

- ~[ ] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~
- [X] _Add tests to cover the fixed bug(s) or the new introduced feature(s) (if appropriate)._
- [X] _Write concise and explanatory changelogs above._
- [X] _If possible, assign one of the following labels to the PR: `feature`, `fix` or `test` (or ask a maintainer to do it for you)._

---


scripts/balance_params_and_train.sh
DomInvivo, line 1: Wasn't that replaced by python code?

graphium/nn/architectures/global_architectures.py
```diff
@@ -1229,15 +1229,15 @@ def create_module_map(self, level: Union[Literal["layers"], Literal["module"]] =
         if self.task_heads is not None:
             self._module_map.update(
                 {
-                    "graph_output_nn/"
+                    "graph_output_nn-"
```
DomInvivo, line 1232: It feels to me that `/` is better since it's less likely to be accidentally mis-used. Why the switch to `-`?

DomInvivo (commented):
Minor comments, can merge as soon as they're answered

graphium/nn/architectures/global_architectures.py
maciej-sypetkowski: I don't know if that's the reason, but hydra doesn't support `/` when overriding from command line (or more precisely `/` token is reserved for another feature). I also had to hack it in the past, so I like the change.

WenkelF: Exactly, that's the reason for the change. Thanks @maciej-sypetkowski

scripts/balance_params_and_train.sh
WenkelF: Yes, but it is not on github. I can remove it if better

DomInvivo: It's fine, we can keep it

DomInvivo (approved):
Approved, feel free to merge when the checks pass

PR #485, opened 2023-10-31
make wandb checkpoint upload optional, change mlp_expansion_ratio def…

kerstink-GC:
…ault from 4 to 2

## Changelogs

---

_Checklist:_

---

### DIFF
--- a/graphium/cli/train_finetune_test.py
+++ b/graphium/cli/train_finetune_test.py
@@ -136,7 +136,9 @@ def run_training_finetuning_testing(cfg: DictConfig) -> None:
     filename_datetime_suffix = now.strftime("%Y%m%d_%H%M%S")
     # Append the datetime string to the existing filename in the cfg dictionary
     cfg["trainer"]["model_checkpoint"]["filename"] += f"_{filename_datetime_suffix}"
-    cfg["trainer"]["model_checkpoint"]["dirpath"] = cfg["trainer"]["model_checkpoint"]["dirpath"][:-1] + f"_{filename_datetime_suffix}"
+    cfg["trainer"]["model_checkpoint"]["dirpath"] = (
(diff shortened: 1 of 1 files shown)


graphium/cli/train_finetune_test.py
```diff
@@ -266,10 +267,11 @@ def run_training_finetuning_testing(cfg: DictConfig) -> None:
     logger.info("Total compute time:", timeit.default_timer() - st)
     logger.info("-" * 50)
 
-    if wandb_cfg is not None:
+    save_checkpoint_to_wandb = cfg["trainer"].get("save_checkpoint_to_wandb")
+    if save_checkpoint_to_wandb is True:
         # Save initial model state - and upload checkpoint to wandb
         if cfg["trainer"]["model_checkpoint"]["save_last"] is True:
-            checkpoint_path = f"{cfg['trainer']['model_checkpoint']['dirpath']}{cfg['trainer']['model_checkpoint']['filename']}.ckpt"
+            checkpoint_path = f"{cfg['trainer']['model_checkpoint']['dirpath']}/{cfg['trainer']['model_checkpoint']['filename']}-v1.ckpt"
```
callumm-graphcore, line 276: Maybe a glob would be better, so we capture all checkpoints?

It's probably worth noting that the `filename` parameter for a W&B parameter can actually be a template, so this might not work if a user does that (ref: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html)

callumm-graphcore:
The title says "change mlp_expansion_ratio default" but I don't see that here?

graphium/cli/train_finetune_test.py
kerstink-GC: So because I added the timestamp to the directory name, there will never be multiple checkpoints in the same folder, so not sure if glob is needed. We could use `last.ckpt` as it seems to be identical to the filename one

PR #494, opened 2023-12-20
Fix mup for the layers with AttentionLayerMup

DomInvivo:
## Changelogs

- Added `embed_dim` to the list of keys to look for when doing the mup kwargs

@maciej-sypetkowski I think this should fix your issue, although I can't verify it because on my end, when I use the config with `architecture.mup_scale_factor: 2` it works. Basically, I can't reproduce because I don't know how you do your scaling for it to fail, but at least the `attn_layer` keys in `mup_base_params.yaml` are no longer `null`.

## IMPORTANT
when this PR closes, it will affect the reproducibility of your models if they use `AttentionLayerMup`, such as `GPSLayerPyg` since the mup will affect the learning rate of these layers, whether they were previously ignored.

### DIFF
--- a/env.yml
+++ b/env.yml
@@ -17,7 +17,6 @@ dependencies:
   - pandas >=1.0
   - scikit-learn
   - fastparquet
-  - networkx
 
   # viz
   - matplotlib >=3.0.1
--- a/env.yml
+++ b/env.yml
@@ -1,6 +1,6 @@
 channels:
   - conda-forge
-  # - pyg # Add for Windows
+  - pyg # Add for Windows
 
 dependencies:
   - python >=3.8
@@ -17,6 +17,7 @@ dependencies:
   - pandas >=1.0
   - scikit-learn
   - fastparquet
+  - networkx
 
   # viz
   - matplotlib >=3.0.1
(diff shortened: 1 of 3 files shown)


graphium/nn/architectures/global_architectures.py
```diff
@@ -1324,6 +1324,12 @@ def _recursive_divide_dim(x: collections.abc.Mapping):
                     _recursive_divide_dim(v)
                 elif k in ["in_dim", "out_dim", "in_dim_edges", "out_dim_edges"]:
                     x[k] = round(v / divide_factor)
+                elif k in ["embed_dim"]:
+                    num_heads = x.get("num_heads", 1)
+                    x[k] = round(v / divide_factor)
+                    assert (
+                        x[k] % num_heads == 0
+                    ), f"embed_dim={x[k]} is not divisible by num_heads={num_heads}"
```
DomInvivo, line 1332: I don't think it's needed since there's another assertion in `AttentionLayerMup`. @maciej-sypetkowski can you check if we remove that part when scaling by a factor that's not divisible by `num_heads`, does it still work?

maciej-sypetkowski: Yes, it's not needed

env.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
   - pandas >=1.0
   - scikit-learn
   - fastparquet
+  - networkx
```
maciej-sypetkowski, line 20: Why is it needed now?

maciej-sypetkowski (approved):
LGTM. Tested and it works

PR #503, opened 2024-01-18
MAJOR-BUG for node/edge predictions: Added ordered=True to `to_mol`

### LINKED ISSUE #467: Wrong ordering of nodes/edges in multitasking for node/edge level tasks
I'll write more details later, but in `MultitaskDataset.merge`, it merges the molecules that are identical across datasets. But if the ordering of the nodes/edges/nodepairs are different, then nothing guarantees that the order will be the same, so the nodes that ends up being used might be in a different orders than the labels. 

In that function, if there is a node/edge/nodepair task, all the labels must be re-ordered according to the ordering of the unique graph selected.

DomInvivo:
A major bug was discovered when working with node, nodepair and edge-level tasks. 

The bug is due to RDKIT's weird behaviour in that it takes the AtomMapping and displays them correctly, but doesn't use them as indices. Unless you specify that you want to canonicalize the molecule. In case you have AtomMaps defined, it will NOT search for a canonical form, instead it will use the AtomMaps.

Below is the weird behaviour by RDKIT, the big numbers next to the carbons are the AtomMapping that are used to order the atoms with the same ordering as the labels. The small numbers are the indices. When visualizing the molecules to validate that the code was working, I didn't not look at the intrinsic indices, but only at the big number, assuming that the indices were correct.

image

**I'm sorry for the inconvenience I have caused everyone...** This bug was discovered this morning when trying to fix #467 .

I have validated the fix experimentally. Simply adding `ordered=True` in `dm.to_mol` yields massive improvements on both the PCQM4M and the PM6 datasets. In the plots below, I randomly subsample 20k molecules and run an MPNN  model on the node-level tasks only. You can see drastic improvements. 

What surprises me is that the metrics on PCQM are not that bad, especially since it can reach r2=0.85 on the 4M large-mix dataset. This led me to believe that the ordering was correct. But as you can see here, we can reach r2=0.95 with only 20k molecules and 100 epochs.

image

### DIFF
--- a/graphium/data/smiles_transform.py
+++ b/graphium/data/smiles_transform.py
@@ -27,7 +27,7 @@ def smiles_to_unique_mol_id(smiles: str) -> Optional[str]:
         mol_id: a string unique ID
     """
     try:
-        mol = dm.to_mol(mol=smiles)
+        mol = dm.to_mol(mol=smiles) # Doesn't need `ordered=True` because the unique_id doesn't depend on the atom order
         mol_id = dm.unique_id(mol)
     except:
         mol_id = ""
--- a/graphium/data/smiles_transform.py
+++ b/graphium/data/smiles_transform.py
@@ -27,7 +27,9 @@ def smiles_to_unique_mol_id(smiles: str) -> Optional[str]:
         mol_id: a string unique ID
     """
     try:
-        mol = dm.to_mol(mol=smiles) # Doesn't need `ordered=True` because the unique_id doesn't depend on the atom order
+        mol = dm.to_mol(
+            mol=smiles
+        )  # Doesn't need `ordered=True` because the unique_id doesn't depend on the atom order
         mol_id = dm.unique_id(mol)
     except:
         mol_id = ""
--- a/graphium/features/featurizer.py
+++ b/graphium/features/featurizer.py
@@ -743,7 +743,7 @@ def mol_to_adj_and_features(
     """
 
     if isinstance(mol, str):
-        mol = dm.to_mol(mol)
+        mol = dm.to_mol(mol, ordered=True)
 
     # Add or remove explicit hydrogens
     if explicit_H:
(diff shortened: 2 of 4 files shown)


s-maddrellmander (changes requested):
- Looking through the rest of the code base we have a number of other places where this pattern is used that haven't been updated, is that intentional or do we need the same logic applied. i.e. is this only needed to be applied to the `mol` objects when using  `dm.to_mol` or with the smiles as well? I've listed a few examples below, but searching the repo gives more. 

 ```python
graphs = []
--
for s in tqdm(smiles):
    mol = dm.to_mol(s)
    graphs.append(mol_to_graph_dict(mol, **featurizer))
 
print(graphs[0])
```

```
mol = dm.to_mol(mol=smiles)
mol_id = dm.unique_id(mol)

```

```python
if isinstance(mol, str):
    mol = dm.to_mol(mol)
```
- Most obviously in the `test_featurizer.py` - which if that needs changing to be ordered we could add an additional check to make sure that molecules are ordered, or if only `mol` objects need ordering a separate check could be added please?
```
err_msg = f"\n\tError for params:\n\t\tSMILES: {s}"
mol = dm.to_mol(s)
```
- I would appreciate it if a test could be added that confirms the ordering and to ensure this can't silently break?

s-maddrellmander (approved):
Cool, thanks for checking!

DomInvivo:
Thanks @s-maddrellmander for reviewing that. I have added comments on why the other calls to `to_mol` don't need `ordered=True`.

In your examples, in order, here are the reasons:

1. This is just some profiling code. It could be added or not, but it's just to test the speed. Doesn't serve a use case.
2. This function only returns the `unique_id`, and this doesn't depend on the atom ordering, only on the graph structure.
3. I added a depreciation warning for all functions in `graphium.features.properties.py` to use `datamol.to_fp` instead. Also, fingerprints don't depend on the node ordering.

Regarding the testing, I have create issue #504 . I currently do not have time to test it properly and believe I should merge the fix, and test it later. I am confident that it will work thanks to the much better training dynamics observed above. If you disagree, and think that it should be tested before being merged, let me know.

PR #521, opened 2024-07-22
Reorder atoms in label data (Fixes 502) + documenting C++

ndickson-nvidia:
## Changelogs

- Reorder atoms in node-level and nodepair-level label data, when the same molecule appears in multiple tasks with different atom orders, to be consistent with the first task's atom order.  This fixes https://github.com/datamol-io/graphium/pull/502
- Document the important C++ functions

---

_Checklist:_

- [x] _Was this PR discussed in an issue? #467 

---


DomInvivo (commented):
I see that canonicalization was used for the re-ordering at dataloading and data preparation time. The only issue is that we can't simply use the "predict mode" for real-world predictions without un-ordering. But I think it's fine. We should probably warn about that in the docs and the code

- [ ] Log warning about canonicalization cannot be used for predictions without un-ordering
- [ ] Warn in the documentation about the point above

PR #522, opened 2024-07-31
Updating dep versions for bh2 install

Andrewq11:
## Changelogs

- Pinning version of `numpy` to the latest version < 2.0. `graphium` uses the old `numpy` API and must be below version 2.0 for now.
- Unpinning the version for the `cuda-version` package. This allows for the CUDA-compatible version of `pytorch` to be installed
- Pinning the version of `rdkit` to a recent version which contains all of the required files (imports) for `graphium_cpp`

---

_Checklist:_

- [ ] ~_Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~
- [ ] ~_Add tests to cover the fixed bug(s) or the new introduced feature(s) (if appropriate)._~
- [ ] ~_Update the API documentation is a new function is added, or an existing one is deleted._~

---

### DIFF
--- a/env.yml
+++ b/env.yml
@@ -12,7 +12,7 @@ dependencies:
   - platformdirs
 
   # scientific
-  - numpy
+  - numpy == 1.26.4
   - scipy >=1.4
   - pandas >=1.0
   - scikit-learn
@@ -28,7 +28,7 @@ dependencies:
   - gcsfs >=2021.6
 
   # ML packages
-  - cuda-version == 11.2 # works also with CPU-only system.
+  - cuda-version # works also with CPU-only system.
   - pytorch >=1.12
   - lightning >=2.0
   - torchmetrics >=0.7.0,<0.11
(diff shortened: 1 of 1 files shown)


env.yml
```diff
@@ -41,7 +41,7 @@ dependencies:
   - pytorch_scatter >=2.0
 
   # chemistry
-  - rdkit
+  - rdkit == 2024.03.4
```
DomInvivo, line 44: Why?? And why not `<= 2024.03.3`?

DomInvivo (commented):
Just a comment about rdkit

env.yml
Andrewq11: The latest version of `rdkit` was not being installed with the required files that are imported by `graphium_cpp`. So, I pinned it to a very recent version. We could do `<= 2024.03.4`, but can we confirm that all earlier versions of `rdkit` contain the required files needed by `graphium_cpp`?

Andrewq11: Let me also clean my `conda` cache and confirm that it wasn't just a local cache issue with the latest version of `rdkit`

Andrewq11: Confirmed that the latest version of `rdkit` on `conda` does not include the required files for `graphium_cpp`

DomInvivo: Yeah this was something that @ndickson-nvidia mentioned as well.

I will merge the PR

PR #525, opened 2024-09-07
Remove ipu

DomInvivo:
## Changelogs

- Removing support for IPU
- Removing custom `to_dense_batch` and `to_sparse_batch` specific to IPUs, and reverted to the pyg version
- Removing all documentation regarding IPUs

---


WenkelF:
@DomInvivo, I have no concerns with the changes. The unit tests might need an overhaul though.

I resolved conflicts in order to run the env checks and they are failing (I believe due to issues with the installation of `graphium_cpp`).

PR #527, opened 2024-09-25
Adding the gcc, gxx, pybind to the env file. Preparing for packaging

DomInvivo:
## Changelogs

- Changing the `env.yml` file to contain all the gcc requirements
- Updating the install instructions

---

_Issues to resolve:_

- [ ] Currently only supports linux x64 because the `gcc` and `gxx` do not work in the env file, we need to specify `gcc_linux-64` and `gxx_linux-64`. Ideally, we want to support Win64, WinARM, Linux64, LinuxARM, Mac64, MacARM
- [ ] Weirdly, `gcc --version` and `gxx --version` don't appear in the command line. We need to rename their folders `ln -s MY_ENV_PATH/bin/x86_64-conda-linux-gnu-gcc MY_ENV_PATH/bin/gcc`, same with `gxx`.
- [ ] When in dev mode, we need to build `graphium/graphium_cpp` separately with `pybind11`, then install it with `pip install -e graphium_cpp --no-deps`. Ideally, all should be built and installed from directly `pip install -e . --no-deps`.

### DIFF
--- a/env.yml
+++ b/env.yml
@@ -49,6 +49,7 @@ dependencies:
   - sympy
   - tensorboard
   - pydantic <2  # because of lightning. See https://github.com/Lightning-AI/lightning/issues/18026 and https://github.com/Lightning-AI/lightning/pull/18022
+  - pytdc
 
   # Dev
   - pytest >=6.0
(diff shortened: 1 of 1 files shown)


DomInvivo:
@Andrewq11 , are you working on that branch? I assume you have your own branch for this, and we can close this one

Andrewq11:
Yes I have my own branch for the packaging work. Feel free to close this.

Andrewq11:
Closing PR as work has evolved on another branch

PR #534, opened 2025-02-07
Fix overzealous label data clipping in save_label_data

ndickson-nvidia:
## Changelogs

- val and test stage data was being clipped to the min and max values of the train stage, even when clipping was not enabled.  In Graphium 2, clipping is only done when both a min and max clipping value are specified in the normalization options, even though those may not be the values used for clipping if the data range of the column doesn't span the whole clipping range.  The behaviour should now match the behaviour of Graphium 2.

---

_Checklist:_

---


DomInvivo (approved):
Thanks for the fix, you can merge

